PR #2, opened 2024-09-29
Init commit

rBergCode:
I just created the basic android project based on [the given android sample](https://github.com/swent-epfl/Android-Sample).
I also updated the `.gitignore` file based on the one given in the bootcamp project


rBergCode:
Did some updates to CI and added `gradle-wrapper.jar`. I check for bootcamp project and it was kind of the same configuration. Everything seems to be okay now.

Wylwi:
Linked to #1

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Mon Mar 11 13:43:48 CET 2024
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
```
Wylwi, line 4: Why not using the latest version of gradle? (8.10.2)

If upgrading is compatible with the project, please run `.\gradlew wrapper --gradle-version 8.10.2 `

app/build.gradle.kts
```diff
@@ -0,0 +1,189 @@
...
+        minSdk = 28
+        targetSdk = 34
+        versionCode = 1
+        versionName = "1.0"
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+        vectorDrawables {
+            useSupportLibrary = true
+        }
+    }
```
Wylwi, line 24: Maybe we should already define those before merging to main?

Wylwi (commented):
Some CI functionalities were missing and need to be added, as well as app configuration. Otherwise, just a standard project initial commit :)

app/build.gradle.kts
rBergCode: I will let them as they are right now until we know the name of the app. We can still change this later

Wylwi (approved):
Everything is good to me. Checks are failing because this is a template, so no issues on my side

PR #5, opened 2024-10-02
chore: update Figma link

mchac15:
Previous link was the project's link, now changed to the template's link


Wylwi (approved):
Link is finally public!

PR #14, opened 2024-10-05
fix(ci): add the jacoco coverage to sonar

rBergCode:
Added the step to generate the jacoco coverage

# Summary
The CI job would not create the Jacoco coverage report. Sonar would then always return a 0.0% coverage.
I added the task to generate the Jacoco coverage report in order for Sonar to watch it.


MaelImhof (approved):
Looks good to me, the Jacoco report should be generated correctly and it seems that the report will be saved to the appropriate location for Sonar to find it. Nice work 👌🏻

Issue #17, opened 2024-10-06
CI not running on branches


Wylwi:
CI isn't running on branches pushes, only if the branch has a pull request open, leading to the necessity of opening a PR each time someone want to check the CI

Wylwi:
Fixed in #18

PR #18, opened 2024-10-06
fix(ci): run ci on all branches

Wylwi:
# Summary
- Fix the #17 issue
- CI wasn't running on all branches, leading to issues when needing to see if Sonar detects issues


mchac15 (approved):
LGTM

PR #23, opened 2024-10-08
feat(auth): Add Firebase backend

davidholzwarth:
# Summary
This PR contains the backend for authentication with Firebase. 

Added files Auth.kt which contain the sign-in and sign-out button as composables, which can be changed according to design preferences, as well as AuthTest.kt, which contains test to test the functionality of the two buttons.

# Details
As discussed, Google's Credential Manager was used instead of Googles ID Sign-in library, which is deprecated. 

The tests for the Authentication are implemented using MockK instead of Mockito, as Mockito does not allow to mock final classes, of which Android's "CredentialManager", which is used in the code, is one.


davidholzwarth:
Note: I closed this PR and will open two separate ones, one for the viewModel and one for the model. This is more in line with the sprint tasks and allows for more a more modular PR.

PR #30, opened 2024-10-11
ci: add a workflow to build an APK

Wylwi:
# Summary

Added the workflow required by the M1 milestone to produce the APK


mchac15 (approved):
LGTM

MaelImhof (approved):
Looks good to me too, the fact that the workflow only runs on manual trigger is a good thing

PR #31, opened 2024-10-11
ci: fix the use of a deprecated action for upload

Wylwi:
# Summary

- Fix use of deprecated actions


MaelImhof (approved):
LGTM, since you ran it locally it should work right? (right?)

Wylwi:
right! right?

PR #36, opened 2024-10-13
feat(map): implement map screen

MaelImhof:
# Summary

This PR introduces a new screen : the map screen.

The map screen features :
- A map where the user can move to see a particular region
- A list of hikes the user can choose from to see more details

The current implementation is only partial, several features still need to be added
- Navigation needs to be integrated
- The view model currently only returns `String`s, it needs to be adapted to return actual `Route`s
- The route details screen must be implemented to have an effect when the user clicks on a hike

Nevertheless, the map is already there and ready to be improved and worked on.

# For the reviewer

Please let me know if you need any help understanding the code I wrote. There are two main files to look into :

- `MapScreen.kt` where the actual map screen is implemented.
    - Contains `object MapScreen` that holds the test tags as constants
    - Contains `fun MapScreen` the composable of the screen itself
    - Contains `fun CollapsibleHikesList` the composable of the list of hikes displayed on top of the map
    - Contains `fun HikingRouteItem` the composable of one hike item in the list
- `MapScreenTest.kt` where tests are implemented. I only implemented very basic tests, because it is hard to test in the details without the ability to mock the repository given to the view model. I added test ideas with `TODO`s inside.

I also edited `strings.xml`. In general, I avoided hardcoding string values and messages in the code directly, I preferred adding them to `strings.xml` for easier translation later and on account of best practices.


app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+) {
+  val context = LocalContext.current
+  // Avoid re-creating the MapView on every recomposition
+  val mapView = remember { MapView(context) }
+
+  Configuration.getInstance().apply {
+    // Set user-agent to avoid rejected requests
+    userAgentValue = context.packageName
+
+    // Allow for faster loading of tiles. Default OSMDroid value is 2.
```
davidholzwarth, line 77: Question: What consequences does increasing this value have for our app?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+    // Maximum number of tiles that can be downloaded at once. Default is 40.
+    tileDownloadMaxQueueSize = 40
+
+    // Maximum number of bytes that can be used by the tile file system cache. Default is 600MB.
+    tileFileSystemCacheMaxBytes = 600L * 1024L * 1024L
+  }
+
+  mapView.apply {
+    controller.setZoom(15.0)
+    // TODO : Set the center to the user's location
```
davidholzwarth, line 89: Nitpick: Perhaps remove the "todo" and replace it with a marker, which you can reference from a task which you create in the product backlog.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+    controller.setCenter(GeoPoint(46.5, 6.6))
+    // Enable touch-controls such as pinch to zoom
+    setMultiTouchControls(true)
+    // Update hiking routes every time the user moves the map
+    // TODO : Those updates could be quite frequent, have a cooldown to avoid sending a request each
```
davidholzwarth, line 94: Note: You can remove this todo, there is a task in the sprint backlog to fix this. Please however add a marker to the code and add it to the task in the sprint backlog, so the person working on the task knows where to look

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+
+          override fun onZoom(event: ZoomEvent?): Boolean {
+            hikingRoutesViewModel.setArea(mapView.boundingBox)
+            return true
+          }
+        })
+  }
+
+  Box(modifier = Modifier.fillMaxSize()) {
+    AndroidView(
```
davidholzwarth, line 111: Nitpick: Perhaps add a comment to explain what AndroidView is

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+        factory = { mapView }, modifier = Modifier.fillMaxSize().testTag(MapScreen.TEST_TAG_MAP))
+
+    IconButton(
```
davidholzwarth, line 114: Nitpick: It would be helpful to have a comment above this button to let the reader know what kind of button this is

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+            Modifier.padding(16.dp)
+                .align(Alignment.TopStart)
+                // Clip needs to be before background
+                .clip(RoundedCornerShape(8.dp))
+                .background(MaterialTheme.colorScheme.surface)
+                .testTag(MapScreen.TEST_TAG_MENU_BUTTON)) {
+          Icon(
+              Icons.Default.Menu,
+              contentDescription =
+                  context.getString(R.string.map_screen_menu_button_content_description),
```
davidholzwarth, line 129: Praise: Nice, I like your usage of constants for this kind of text so that our app is more consistent.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+              tint = MaterialTheme.colorScheme.onSurface)
+        }
+
+    CollapsibleHikesList(hikingRoutesViewModel)
+  }
+
+  // Initialize the list of hiking routes once when the map is loaded
+  // Leave subsequent updates to the map listener
```
davidholzwarth, line 137: Question: What does this mean? Does the Launched Effect take over? I am not sure what this comments means

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun CollapsibleHikesList(hikingRoutesViewModel: ListOfHikeRoutesViewModel) {
+  val scaffoldState = rememberBottomSheetScaffoldState()
+  val routes = hikingRoutesViewModel.hikeRoutes.collectAsState()
+  val context = LocalContext.current
+
+  BottomSheetScaffold(
```
davidholzwarth, line 148: Nitpick: Here also a comment would be helpful, explaining what BottomSheetScaffold does

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+                      text = context.getString(R.string.map_screen_empty_hikes_list_message),
+                      style = MaterialTheme.typography.bodyLarge,
+                      // Align the text within the Text composable to the center
+                      textAlign = TextAlign.Center)
+                }
+              }
+            } else {
+              items(routes.value.size) { index: Int ->
+                HikingRouteItem(
+                    title = routes.value[index],
```
davidholzwarth, line 168: Note: Consider add a comment here that there values are hardcoded and will be updated with real values later

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+                    altitudeDifference = 1000,
+                    isSuitable = index % 2 == 0,
+                    onClick = {
+                      // TODO : Navigate to the details of the selected hiking route
```
davidholzwarth, line 172: Warning: Another todo here, make sure to not leave this in the code and rather add tasks to the backlog

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+                      Toast.makeText(
+                              context, "Hike details not implemented yet", Toast.LENGTH_SHORT)
+                          .show()
+                    })
+              }
+            }
+          }
+        }
+      },
+      sheetPeekHeight = 400.dp) {}
```
davidholzwarth, line 182: Question: Does this mean the max height that you can pull the collapsible menu? Is there a reason you hardcoded this value to 400 dp? It seems to me that either using a relative measure like 3/5 of the screen height or just completely not restricting the height might make more sense.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+                  Text(
+                      text = context.getString(R.string.map_screen_empty_hikes_list_message),
+                      style = MaterialTheme.typography.bodyLarge,
+                      // Align the text within the Text composable to the center
+                      textAlign = TextAlign.Center)
+                }
+              }
+            } else {
+              items(routes.value.size) { index: Int ->
+                HikingRouteItem(
```
davidholzwarth, line 167: Praise: great modularity :), make this code very readable

app/src/main/res/values/strings.xml
```diff
@@ -2,4 +2,19 @@
     <string name="app_name">SampleApp</string>
     <string name="title_activity_main">MainActivity</string>
     <string name="title_activity_second">SecondActivity</string>
+
+    <!-- Map Screen -->
```
davidholzwarth, line 6: Praise: Love the use of constants for text in the UI!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+
+          Spacer(modifier = Modifier.height(4.dp))
+
+          Row(verticalAlignment = Alignment.CenterVertically) {
+            Icon(
+                imageVector = if (isSuitable) Icons.Default.Check else Icons.Default.Warning,
+                // The icon is only decorative, the following message is enough for accessibility
+                contentDescription = null,
+                // TODO : Replace suitable and challenging icon colors with theme colors
+                tint = if (isSuitable) Color(0xFF4CAF50) else Color(0xFFFFC107),
```
davidholzwarth, line 239: Nitpick: If you do decide to remove todos and replace them in with issues, afaik there is already a task in the product backlog on adding a theme file, you could just add to this task to replace the colors in this section here.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+            Text(
+                text =
+                    if (isSuitable)
+                        LocalContext.current.getString(R.string.map_screen_suitable_hike_label)
+                    else LocalContext.current.getString(R.string.map_screen_challenging_hike_label),
+                style = MaterialTheme.typography.bodySmall,
+                // TODO : Replace suitable and challenging icon colors with theme colors
+                color = if (isSuitable) Color(0xFF4CAF50) else Color(0xFFFFC107))
+          }
+        }
```
davidholzwarth, line 251: Praise: I like your treatment of insuitable hikes, changing the colors, adding a warning symbol and changing the text!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+
+        Spacer(modifier = Modifier.width(4.dp))
+
+        Icon(
```
davidholzwarth, line 255: Question: What does this Icon do? Consider adding a comment

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
```diff
@@ -0,0 +1,58 @@
+package ch.hikemate.app.ui.map
```
davidholzwarth, line 1: Request: Please let me know once you've pushed the test so that I can review them!

davidholzwarth (commented):
Great work, a lot of the comments are nitpicks or requests for clarification via comments. :)

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Thank you !

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
MaelImhof: The tests that are currently implemented will be the only ones for this PR. I will implement most of the other ones when working on #41, because I need to perform some modification of how the view model works before testing the map screen.

I can ask you for a review on the other PR once it will be ready though.

If that is OK to you, we can mark this conversation as resolved and I'll request your review on the new PR I'll create for #41.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: I asked ChatGPT this :

```
Hey there, I'm building an Android app in Kotlin using Jetpack compose and OSMDroid, and I got interested with the configuration option tileDownloadThreads of OSMDroid.

My manager asked me : What consequences does increasing this value have for our app?
```

(Yes, you are my manager xD)

And did some research on what it said.

In summary, the response is : it is a tradeoff, mainly depending on two factors
1. **The user device :** for old devices, a higher number of threads would mean a higher performance overhead, but on newer devices the overhead is negligible compared to the increase in speed it gives. More threads means more tiles downloaded in parallel.
2. **The user context :** in particular the network they are using. If the network is slow (for example in the mountains), having multiple threads might have the inverse effect and be a problem for performance, leading to congestion of the network and more packet loss. On the other hand, on a good network, downloading in parallel is faster.

In the end, it's an experimental value to play around with. Since Hike Mate is not destined to very old phones (at least that's what I understood), increasing it to 4 seems fine to me. We might even increase it further, but 4 already gives a pretty decent experience.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: I'm not sure about what you mean in "replace it with a marker", what is a marker exactly? Should I leave a comment but simply remove the TODO and use another keyword?

MaelImhof: I think it would be cleaner to only have everything written once in the backlog item. If we have two places where the task is mentioned, we won't think about updating both in case we do update the task.

I could simply mention the line or lines to change directly in the issue or the backlog item, such as :

```kotlin
mapView.apply {

    // Other config options

    // TODO : Update this line to center the map on the user's location
    controller.setCenter(GeoPoint(46.5, 6.6))
    
    // Other config options
  }
```

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Might not be so obvious indeed. Would this comment be a good explanation?

```kotlin
// Jetpack Compose is a relatively recent framework for implementing Android UIs. OSMDroid is
    // an older library that uses Activities, the previous way of doing. The composable AndroidView
    // allows us to use OSMDroid's legacy MapView in a Jetpack Compose layout.
```

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Will do in a commit where I add comments you suggested 👌🏻

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: I guess, even though defining what `BottomSheetScaffold` is in a few words is difficult. I'm not very good at summarizing lol. I'm open to any suggestions, I thought about the following :

```kotlin
// BottomSheetScaffold adds a layout at the bottom of the screen that the user can expand. This
// is ideal for showing a list of items on top of a map. The user can either expand the list and
// only view the hikes list, or collapse it and see the map.
```

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Thank you that's great to hear !

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: I created [this draft task](https://github.com/orgs/HikeMate/projects/1/views/2?pane=issue&itemId=83415053) in the backlog, let me know what you think

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Actually this TODO is irrelevant based on what we discussed about #41. Since there will be a _Search Here_ button, those events do not need to be throttled anymore, because they won't be used.

I'll remove the TODO altogether. I added information about those event listeners in #41 directly (even though I'm assigned to it).

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Added this :

```kotlin
// Burger menu button in the top left corner of the screen
```

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: I created [this draft task](https://github.com/orgs/HikeMate/projects/1/views/2?pane=issue&itemId=83421220) for the altitude difference. `isSuitable` would require more work from what I gather, it would rather be a user story

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Fair, however this task would be a lot of elements :
- Create the Hike details screen in Figma
- Implement the Hike details screen -> Implement navigation from map screen to hike details screen

So I'm not sure I should create all of those tasks already, maybe keeping the TODO is better.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: The maximum height is : fill screen. The value of `400.dp` represents the _minimum_ height. I just played around with it and `400.dp` looked good, that's pretty much it.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Done 👌🏻

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Thanks !

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Will do, added this :

```kotlin
// Arrow icon to indicate that the item is clickable
```

app/src/main/res/values/strings.xml
MaelImhof: Thanks I appreciate it 😄

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Thanks for researching! I agree that in this case 4 is fine for the moment, and if we notice the map loads slowly or we have issues with we can adjust this value.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: This looks great, thanks!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Yes, that makes it clear.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Thanks for clarifying, the updated comment is good!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Fair enough, imo even just this suffices: `// BottomSheetScaffold adds a layout at the bottom of the screen that the user can expand to view the list of hikes.`

MaelImhof: Will use yours instead, a shorter comment will be better.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: I'm afraid it might get lost in the code, and a task would ensure oversight, and ensure that all our upcoming tasks are in one place. Perhaps its enough to summarise all these tasks in one task for now, and the SCRUM master can figure out how to best split them up?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Oooh, 'peek' height, not 'peak' height, I'm sorry haha. In that case this should be fine.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
MaelImhof: Indeed, putting all of the tasks in one place is the best option. I'll create the tasks now.

MaelImhof: Created [this draft task](https://github.com/orgs/HikeMate/projects/1/views/2?pane=issue&itemId=83430191), what do you think @davidholzwarth?

app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,19 @@
...
+    <string name="map_screen_suitable_hike_label">
+        This hike is suitable for your level
+    </string>
+    <string name="map_screen_challenging_hike_label">
+        This hike would be an interesting challenge
+    </string>
+    <string name="map_screen_hike_details_content_description">
+        Go to hike details
+    </string>
 </resources>
```
davidholzwarth, line 19: Thanks for including my changes!

davidholzwarth (approved):

PR #44, opened 2024-10-14
ci: add signing the APK to the workflow

Wylwi:
# Summary

- Changed the gradle configuration to sign the APK on a release

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,12 @@ jobs:
       - name: Decode secrets
         env:
           GOOGLE_SERVICES: ${{ secrets.GOOGLE_SERVICES }}
+          KEYSTORE_PROPERTIES: ${{ secrets.KEYSTORE_PROPERTIES }}
+          KEYSTORE: ${{ secrets.KEYSTORE }}
         run: |
           echo "$GOOGLE_SERVICES" | base64 --decode > ./app/google-services.json
+          echo "$KEYSTORE_PROPERTIES" | base64 --decode > ./keystore.properties
+          echo "$KEYSTORE" | base64 --decode > ./keystore.jks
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -31,8 +31,12 @@ jobs:
       - name: Decode secrets
         env:
           GOOGLE_SERVICES: ${{ secrets.GOOGLE_SERVICES }}
+          KEYSTORE_PROPERTIES: ${{ secrets.KEYSTORE_PROPERTIES }}
+          KEYSTORE: ${{ secrets.KEYSTORE }}
         run: |
           echo "$GOOGLE_SERVICES" | base64 --decode > ./app/google-services.json
+          echo "$KEYSTORE_PROPERTIES" | base64 --decode > ./keystore.properties
+          echo "$KEYSTORE" | base64 --decode > ./keystore.jks
```
rBergCode, line 39: Question: Why would you need that for the CI ? Shouldn't that only be required for building the APK, and so, only in the `artifact_building.yml` ?

rBergCode (approved):
LGTM. I only have a quick question in a comment.

.github/workflows/ci.yml
Wylwi: Since loading the keystore.properties is done in the top level section of the build.gradle.kts, it is done regardless of the build currently going on. So, without the file, it fails (see [failing action](https://github.com/HikeMate/hikeMateApp/actions/runs/11335332315/job/31523270277)). I will look into it to see if it's possible to move it or to put conditions, but the Android tutorial was just showing this way, so I did it this way

Wylwi: Looked about it. Gradle wants to load the file, even if the keystore isn't used, so I will leave this as it.

Issue #38, opened 2024-10-14
Clean up project


Wylwi:
- [x] Remove Template useless files, check that nothing related to "Sample app" is left
    - [x] Change the name of the project to _Hike Mate_ so that the Android Studio project has title "Hike Mate" instead of "SampleApp"
- [x] Add theme from Figma (Font, colors, font sizes, etc.)
- [x] Clean up app/build.gradle.kts
- [x] Change sonar rule to 80% code coverage on all code, not only new code
- [x] Multiple Junit in libs
- [x] Multiple sonar tasks in CI
- [x] Similar step CI name ("Run tests" and "run tests")
- [x] Fix Sonar issues
- [ ] List issues that should not be reproduced somewhere so they get added to the wiki (adding them to the wiki itself is not part of the task)

Wylwi:
Closed by #42

PR #37, opened 2024-10-14
Feat/firebase authentication repository

davidholzwarth:
# Summary

Add a repository that can authenticate users using firebase. Currently, users can only sign in using their Google account. This repository will be followed by a viewModel, so the UI can easily access the functionality in this repository.

# Details

This pull request introduces a new method signInWithGoogle() within the FirebaseAuthRepository class to handle authentication with Firebase using Google Sign-In. The method integrates the Android CredentialManager to request user credentials and uses the Firebase SDK to authenticate via Google credentials.

**Functionality:**
signInWithGoogle():

Authenticates users via Google accounts already signed in on the device using Firebase Authentication.
Utilizes Android's CredentialManager to request credentials, streamlining the sign-in process.
Incorporates coroutines to handle asynchronous tasks, avoiding UI thread blocking.
Handles success and error callbacks via parameters (onSuccess, onErrorAction).

signOut():

Signs out the current user

PR #45, opened 2024-10-15
Feat/firebase authentication view model

davidholzwarth:
# Summary

Added the viewModel for authentication with Firebase and corresponding tests.

# Details

Added two methods for the UI to use:
- `signInWithGoogle`
Uses the current context and a coroutine scope, which get passed as parameters. The method goes on to call the `FirebaseAuthRepository`, which provides the backend for the authentication. The viewModel has no onSuccess parameter or return value, but rather updates the state of "currentUser". This means that any UI will need to constantly check for changes in the state of this stateFlow.

- `signOut`
Clears the `currentUser`, effectively signing out the user. This again has no return value but rather changes the state of the `currentUser`stateFlow.


rBergCode (changes requested):
Hey!
It looks like you have not rebase from main before pushing. Specifically #38 which cleans a lot of the project.
Also, please wait for #37 to be accepted, then merge it to main, then also rebase from main so that there are not the repository changes in this branch.

**TL;DR**: #37 should be merged to main and rebased to this branch before I can review this code.

Thanks!

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
```diff
@@ -0,0 +1,144 @@
...
+  private fun setupSignedInUser() {
+    `when`(mockFirebaseAuth.currentUser).thenReturn(mockFirebaseUser)
+    viewModel = AuthViewModel(mockRepository)
+  }
+
+  // Helper functions to setup the ViewModel with a signed-in or signed-out user
+  private fun setupSignedOutUser() {
+    `when`(mockFirebaseAuth.currentUser).thenReturn(null)
+    viewModel = AuthViewModel(mockRepository)
+  }
```
rBergCode, line 67: Nice! Love how you did that.

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
```diff
@@ -0,0 +1,144 @@
...
+          null
+        }
+        .`when`(mockRepository)
+        .signInWithGoogle(any(), any(), any(), any(), any())
+
+    // Verify that currentUser is initially null
+    assertNull(viewModel.currentUser.first())
+
+    viewModel.signInWithGoogle(this, mockContext)
+
```
rBergCode, line 114: I think you forgot to check if the function calls the repository here. You did it in the sign out test.
I suppose you should add something like `verify(mockRepository).signInWithGoogle(any(),any(),any(),any(),any())` or something like that.

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
```diff
@@ -0,0 +1,144 @@
...
```
rBergCode, line 92: I think you forgot to check if the function calls the repository here. You did it in the sign out test.
I suppose you should add something like `verify(mockRepository).signInWithGoogle(any(),any(),any(),any(),any())` or something like that.

rBergCode (approved):
Nice job!

PR #48, opened 2024-10-16
feat(map): show hikes on the map

rBergCode:
Added a function to show a hike on the map using a Polyline.

# Summary
The objective was to show a mock hike on the map as a line. For now, I display 3 mock routes on the map.
There are no tests on the `onClick` of the line as there are no ways to retrieve a line on the map (no test tag for example)

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
@@ -34,6 +34,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.font.FontWeight
@@ -42,6 +43,9 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
 import androidx.lifecycle.viewmodel.compose.viewModel
 import ch.hikemate.app.R
+import ch.hikemate.app.model.route.Bounds
+import ch.hikemate.app.model.route.HikeRoute
+import ch.hikemate.app.model.route.LatLong
 import ch.hikemate.app.model.route.ListOfHikeRoutesViewModel
 import ch.hikemate.app.ui.navigation.LIST_TOP_LEVEL_DESTINATIONS
 import ch.hikemate.app.ui.navigation.NavigationActions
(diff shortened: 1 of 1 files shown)


app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -45,19 +46,80 @@ import androidx.compose.ui.unit.dp
...
+
+  const val STROKE_WIDTH = 10f
+}
+
+/**
+ * Generates a random color in the HSL color space. This function uses the HSL color space to
+ * generate a random color as the HSL is better to generate more visually appealing colors.
+ *
+ * @return The generated color as an [Int].
+ */
```
MaelImhof, line 82: Suggestion: I don't find this documentation very clear. If it wasn't for your IRL explanation, I wouldn't necessarily have gotten it. Maybe mention the fact that HSL has attributes that better represent whether a color will go well on the map or not. For example :

```suggestion
/**
 * Generates a random color in the HSL color space. The function uses HSL to generate a
 * color instead of ARGB because it makes it easier to have a range of colors that we know
 * will render well on the map.
 *
 * @return The generated color as an [Int].
 */
```

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -45,19 +46,80 @@ import androidx.compose.ui.unit.dp
...
+fun getRandomColor(): Int {
+  fun randomInt(min: Int, max: Int): Int {
+    return (min..max).random()
+  }
```
MaelImhof, line 86: Question: is this better than or equivalent to

```kotlin
import kotlin.random.Random

// ...

  fun randomInt(min: Int, max: Int): Int {
    return Random.nextInt(min, max + 1)
  }
```

I mean it does not seem to be a problem so we can keep your version, I think it's more elegant, I was just surprised by the implementation.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -45,19 +46,80 @@ import androidx.compose.ui.unit.dp
...
+  line.outlinePaint.strokeWidth = MapScreen.STROKE_WIDTH
+
+  line.setOnClickListener { _, _, _ ->
+    Toast.makeText(
+            mapView.context,
+            "Hike details not implemented yet. Hike ID: ${hike.id}",
+            Toast.LENGTH_SHORT)
+        .show()
+    true
+  }
```
MaelImhof, line 120: Praise: very elegant method, I like it !

Just to mention, I added this `setOnClickListener` to the backlog so that it gets updated when we implement the Hike details screen.

MaelImhof (approved):
Nice work ! My only suggestion would be to update the docstring of `getRandomColor` because it is not very clear to me. Other than that, ready to be merged.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: Love your suggestion! Changed it

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: From my research, both do the same job. From what I understood, your version is a bit more performant as it will not create a range (which is very minimal). The following method would be call in my version:
```kotlin
@SinceKotlin("1.3")
public fun Random.nextInt(range: IntRange): Int = when {
    range.isEmpty() -> throw IllegalArgumentException("Cannot get random in empty range: $range")
    range.last < Int.MAX_VALUE -> nextInt(range.first, range.last + 1)
    range.first > Int.MIN_VALUE -> nextInt(range.first - 1, range.last) + 1
    else -> nextInt()
}
``` 
Which would use your version.

If it's good to you, I will keep my version as it is, as you said, more elegant.

MaelImhof (approved):
Great, looks ready to me !

PR #49, opened 2024-10-16
fix: fix handling of alternative routes in overpass API

Wylwi:
# Summary 
- The Overpass API can return alternative routes, which are currently not supported, but were incorrectly handled

# Details
- The parser was skipping the name because the value was already read, thus leading to an error on next iteration


rBergCode (approved):
LGTM! Nice regression test!

PR #53, opened 2024-10-17
feat: merge the firebase view model with the sign in screen

Wylwi:
# Summary

- The goal was to merge the created ViewModel for Firebase authentication with the UI (The SignInScreen)

# Details

- Some adaptations were necessary in the ViewModel. I introduced a Factory to make it easier and cleaner to be built
- Tests which were checking if the navigation was going from one screen to another is not relevant since the google sign in introduce a call to another activity (Nevertheless, if you have an idea on how to cleanly test this, let me know)
- I added a single feedback (with a Toast) in case the Google Log in fails because of no account in the device


app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -61,6 +65,8 @@ class FirebaseAuthRepository {
                 request = request, // Send the request we built
                 context = context // Provide the context for the request
                 )
+        if (result.credential.type != GoogleIdTokenCredential.TYPE_GOOGLE_ID_TOKEN_CREDENTIAL)
+            throw Exception("Invalid credential type")
```
rBergCode, line 69: Could you document why you did this here ? Was there an issue without this check ?

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -80,6 +86,13 @@ class FirebaseAuthRepository {
             onErrorAction(task.exception ?: Exception("Unknown error"))
           }
         }
+      } catch (e: NoCredentialException) {
+        // In case no credentials are found, show a toast message and invoke the error callback
+        Log.d("SignInButton", "No credentials found: ${e.message}")
+        CoroutineScope(Dispatchers.Main).launch {
+          Toast.makeText(context, "No Google account found on the device", Toast.LENGTH_LONG).show()
+        }
+        onErrorAction(e)
```
rBergCode, line 95: I like the way you handle the error. But I think display a Toast should not be in the repository as this is related to UI.
I don't think you can do it in the Screen as of how the architecture is but you could do it in the view model.

app/src/test/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
```diff
@@ -179,6 +184,56 @@ class FirebaseAuthRepositoryTest {
...
+            coroutineScope = this)
+
+        advanceUntilIdle()
+
+        coVerify { mockCredentialManager.getCredential(any(), any<GetCredentialRequest>()) }
+        verify(exactly = 0) { mockFirebaseAuth.signInWithCredential(mockAuthCredential) }
+        verify(exactly = 0) { onSuccess(any()) }
+        verify { onError(any<NoCredentialException>()) }
+      }
+
```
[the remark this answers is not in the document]
Wylwi: Thanks!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
Wylwi: Yep, I fully agree, I'm editing it

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
Wylwi: Done :)

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
rBergCode: Also, is this the official way to handle it ? By that I mean, is it mandatory to have a Google account link to the device to be able to sign in with Google ?

rBergCode (changes requested):
Nice job! I still have some questions and suggestions you can see in the comments before merging.

rBergCode (approved):
LGTM! Thank you for having moved the Toast in the Screen. Looks better!

rBergCode (dismissed):
LGTM!

rBergCode:
We close it as it was done in #93.

PR #55, opened 2024-10-17
test(end2end): first end2end test with basic nav

rBergCode:
# Summary
The objective of this PR is to create a full and complete **end to end test** in order to test the app as a real user.

This is done using `composeTestRule` and checking for `TEST_TAGs` throughout the app.


rBergCode:
As of right now, we won't have much time to do more in this end2end test. It still tests the whole navigation.

PR #58, opened 2024-10-19
SwEnt Milestone 1 Feedback

Andy130604:
Use this mechanism to make M1 feedback reports available to teams to whose repos we don't have suitable access


Wylwi (approved):
LGTM

PR #59, opened 2024-10-20
Fix/firebase credential error

davidholzwarth:
# Summary

This PR fixes the "Missing Credential" issue encountered in the Authentication process when the user has never added a Google account to the android device.

# Details

The `NoCredentialException` error was caused by the edge case of a phone having never had a google account connected to it. If at least one account is connected, the error does not happen. 

To fix this, a catch was implemented that catches this error and launches an activity thats prompts the user to add an account via a `addGoogleAccountIntent` activity. 

*Note*: Currently after this has happened successfully, the login still fails, since the login has failed once the error `NoCredentialException` is thrown, and Kotlin's Lifecycle Awareness does not allow for this method to get re-call itself in a simple way after the `addGoogleAccountIntent` activity has finished. However, in the UI the passed parameter `addAccountLauncher`, which gets passed to `firebaseAuthRepository.signInWithGoogle` via `authViewModel.signInWithGoogle`, can be created in such a way that once the `addGoogleAccountIntent` has finished a further call to `authViewModel.signInWithGoogle` is triggered and the login gets re-tried. This will be done in another PR.

The repository not retriggering the sign in by itself is not satisfying and there might exist a cleaner way which I don't have the time to look for tonight. Perhaps in a future PR we could try coding this functionality in a more satisfying way.


app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -33,9 +39,10 @@ class FirebaseAuthRepository {
       context: Context,
       coroutineScope: CoroutineScope,
       credentialManager: CredentialManager = CredentialManager.create(context),
+      startAddAccountIntentLauncher: ManagedActivityResultLauncher<Intent, ActivityResult>?,
```
Wylwi, line 42: Question: I haven't the code for the View, but is it necessary to pass the ActivityLauncher from the View to the Model? Isn't there any way to get it in the ViewModel or the Model from the context?

Wylwi, line 42: Question: Why is the `startAddAccountIntentLauncher` nullable?

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
```diff
@@ -25,12 +28,19 @@ class AuthViewModel(private val repository: FirebaseAuthRepository) {
    * @param coroutineScope The CoroutineScope in which this operation will be executed.
    * @param context The context is used to start the Google Sign-In process.
    */
-  fun signInWithGoogle(coroutineScope: CoroutineScope, context: Context) {
+  fun signInWithGoogle(
+      coroutineScope: CoroutineScope,
+      context: Context,
+      startAddAccountIntentLauncher: ManagedActivityResultLauncher<Intent, ActivityResult>?,
```
Wylwi, line 34: Question: Why is the `startAddAccountIntentLauncher` nullable?

app/src/androidTest/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
```diff
@@ -161,22 +166,51 @@ class FirebaseAuthRepositoryTest {
...
+            onErrorAction = mockOnError,
+            context = context,
+            credentialManager = mockCredentialManager,
+            coroutineScope = this,
+            startAddAccountIntentLauncher = mockLauncher)
+
+        advanceUntilIdle()
+
+        // Verify that getCredential() was called and threw the NoCredentialException
+        coVerify { mockCredentialManager.getCredential(any(), any<GetCredentialRequest>()) }
```
Wylwi, line 206: Nitpick: This check only checks if getCredential() is called, but not for exception throwing

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -81,12 +88,24 @@ class FirebaseAuthRepository {
           }
         }
       } catch (e: Exception) {
-        Log.d("SignInButton", "Login error: ${e.message}")
-        onErrorAction(e)
+        // This check is necessary bc the type of exception is always UndeclaredThrowableException
+        if (e.cause is NoCredentialException) {
```
Wylwi, line 92: Fix: I've tried running your code, It doesn't work when no credentials are available. The issue is this line, since e.cause is a Throwable and not an Exception. Implementing the below described multiple catch clause is cleaner and should fix this issue too

Wylwi (changes requested):
Just a few fixes to implement, since the code doesn't work when there is no account on the device. I've made some suggestions about it! It shouldn't take too long. It should be ready for merging after this!

app/src/androidTest/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
davidholzwarth: You're right, and its not supposed to check for the throwing. I amended the comment.

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
davidholzwarth: I wrote a possible implementation of the launcher while testing my functionality, and it looked something like this:
```kotlin val addAccountLauncher =

          rememberLauncherForActivityResult(
            contract = ActivityResultContracts.StartActivityForResult()) {
              result: ActivityResult ->
            authViewModel.signInWithGoogle(coroutineScope, context, null)
            Toast.makeText(context, "success", Toast.LENGTH_SHORT).show()
            Log.d("MainActivity", "addAccountLauncher result: $result")
          }
```

This `rememberLauncherForActivityResult` is only callable from a composable, so in the UI. This is why I decided to pass it all the way down... If you know a nicer way to solve this I'd be happy to hear, I also don't love this solution.

Wylwi: Yep, okay, I see, but I think you answered the wrong comment!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
davidholzwarth: I wrote a possible implementation of the launcher while testing my functionality, and it looked something like this:
```kotlin 
val addAccountLauncher =
          rememberLauncherForActivityResult(
            contract = ActivityResultContracts.StartActivityForResult()) {
              result: ActivityResult ->
            authViewModel.signInWithGoogle(coroutineScope, context, null)
            Toast.makeText(context, "success", Toast.LENGTH_SHORT).show()
            Log.d("MainActivity", "addAccountLauncher result: $result")
          }
```

This `rememberLauncherForActivityResult` is only callable from a composable, so in the UI. This is why I decided to pass it all the way down... If you know a nicer way to solve this I'd be happy to hear, I also don't love this solution.

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
davidholzwarth: Ah, my bad, the launcher is nullable since after connecting the account together was successful, we make the call to the authViewModel that I also show in my answer. 
```kotlin 
authViewModel.signInWithGoogle(coroutineScope, context, null)`
If we the launcher was not nullable, we would have to pass the addAccountLauncher itself as parameter here, which does not work afaik. It being null is not an issue however, since the NoCredentialError should be resolved after the google sign in is completed, so the addAccountLauncher, which is only ever used when a NoCredentialError is catched, should not need to be called.

File: app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt (hunk 1 above)
Comment (davidholzwarth) [reply app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt:42]: The launcher is nullable since after connecting the account together was successful, we make the call to the authViewModel that I also show in my answer. 
```kotlin 
authViewModel.signInWithGoogle(coroutineScope, context, null)
```

If the launcher was not nullable, we would have to pass the addAccountLauncher itself as parameter here, which does not work afaik. It being null is not an issue however, since the NoCredentialError should be resolved after the google sign in is completed, so the addAccountLauncher, which is only ever used when a NoCredentialError is catched, should not need to be called.

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -81,12 +88,24 @@ class FirebaseAuthRepository {
...
+          // If there is no Google account connected to the device, prompt the user to add one
+          startAddAccountIntentLauncher?.launch(getAddGoogleAccountIntent())
+        } else {
+          Log.d("SignInButton", "Login error: ${e.message}")
+          onErrorAction(e)
+        }
```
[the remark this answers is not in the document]
davidholzwarth: Yes, I used to have it like this before I wrote tests, however when I mocked the mockCredentialManager's .getCredential function to throw a NoCredentialException it always threw an UndeclaredThrowableException, and this was the way to catch such an exception.

I changed it back now and removed the test completely.

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
davidholzwarth: I changed it back now and it should work now, thanks!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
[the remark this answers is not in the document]
Wylwi: Mmmh, looks kinda strange but thanks!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
davidholzwarth: If the google sign in failed, this would result in the second call to the authViewModel to fail also, so ultimately just nothing will have happened. For example, if there is no internet, then the account connection will fail, and subsequently so will the second call to the authViewModel, and the user is just back to the sign in screen.

In that case the user can just click the sign in button again and retry.

Wylwi: Oh okay thanks!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
Wylwi: Thanks!

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
Wylwi: Mmmmh, I see. Definitely not ideal, but I don't see any way to improve it without a whole refactor 😅. So leave it like it

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
Wylwi: Thanks for the detailed explanation!

Wylwi (approved):
Nice patch! Thanks for the modifications and library cleanup! LGTM!

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
```diff
@@ -10,7 +10,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 
-class AuthViewModel(private val repository: AuthRepository) : ViewModel() {
+class AuthViewModel(private val repository: FirebaseAuthRepository) {
```
rBergCode, line 13: Warning: I think you missed merge this. You should keep the Interface version (with the `ViewModel()`)

app/src/androidTest/java/ch/hikemate/app/ui/profile/ProfileScreenTest.kt
```diff
@@ -32,7 +32,7 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 
 class ProfileScreenTest : TestCase() {
-  private lateinit var authRepository: AuthRepository
+  private lateinit var authRepository: FirebaseAuthRepository
```
rBergCode, line 35: Warning: I think you missed merge here. You should keep the `authRepository: AuthRepository`

app/src/androidTest/java/ch/hikemate/app/ui/profile/ProfileScreenTest.kt
```diff
@@ -47,7 +47,7 @@ class ProfileScreenTest : TestCase() {
     context = ApplicationProvider.getApplicationContext()
 
     navigationActions = mock(NavigationActions::class.java)
-    authRepository = mock(AuthRepository::class.java)
+    authRepository = mock(FirebaseAuthRepository::class.java)
```
rBergCode, line 50: Same remark here

app/build.gradle.kts
```diff
@@ -181,26 +181,33 @@ dependencies {
...
+    androidTestImplementation(libs.compose.test.junit)
+    androidTestImplementation(libs.mockito.android)
+    androidTestImplementation(libs.mockk.android)
+    androidTestImplementation(libs.mockk.agent)
+
 
     // Debug dependencies
     debugImplementation(libs.compose.tooling)
     debugImplementation(libs.compose.test.manifest)
+    debugImplementation(libs.compose.test.manifest)
```
rBergCode, line 210: Warning: You have a duplicate line here

app/build.gradle.kts
```diff
@@ -181,26 +181,33 @@ dependencies {
...
+    testImplementation(libs.mockito.core)
+    testImplementation(libs.mockito.inline)
+    testImplementation(libs.mockk.android.v10)
+    testImplementation(libs.mockk.agent.v10)
 
     // Android test dependencies
     androidTestImplementation(libs.androidx.espresso.core)
-    androidTestImplementation(libs.mockito.android)
     androidTestImplementation(libs.androidx.junit)
+    androidTestImplementation(libs.compose.test.junit)
```
rBergCode, line 199: Warning: You have this in duplicate

app/build.gradle.kts
```diff
@@ -181,26 +181,33 @@ dependencies {
     globalTestImplementation(libs.compose.test.junit)
     globalTestImplementation(libs.kaspresso)
     globalTestImplementation(libs.kaspresso.compose)
-    globalTestImplementation(libs.junit)
```
rBergCode: Warning: I've seen you put `libs.junit` in both test and androidTest implementations. You can keep it to global then

app/build.gradle.kts
```diff
@@ -181,26 +181,33 @@ dependencies {
...
-    globalTestImplementation(libs.compose.test.junit)
```
rBergCode: Same remark as for libs.junit

rBergCode (changes requested):
I think you missed merge some of the changes. After correcting the comments you should be good to go!

rBergCode (approved):
LGTM thanks!

Issue #20, opened 2024-10-07
Integrate the Open Street Map API for the map


Wylwi:
Get used to the API, create some functions to use it, etc.

When writing tests, see whether [avoiding hard-coded dispatchers](https://sonarcloud.io/project/issues?pullRequest=13&open=AZJikO6Fec0t_iyog9lw&id=HikeMate_hikeMateApp) for coroutines would be an advantage.

PR #62, opened 2024-10-21
feat: incorporate the Open Elevation API into the project to get the elevation of a certain coordinate

mchac15:
# Summary
This pull request introduces the `ElevationServiceRepository` class to handle elevation requests using the Open Elevation API. It also enhances the testing suite by adding a long list of test coordinates and a comprehensive mock response for elevation data.

# Changes Made
- **ElevationServiceRepository**:
  - Implemented the `getElevation` function to send elevation requests with coordinates.
  - Utilized `OkHttp` for network calls, handling both successful and failed responses.
  - Parsed elevation results from the JSON response using serializable data classes.

# Features
- The ability to fetch elevations for multiple coordinates in a single request.
- A cache feature could be added by caching the elevation of the points of a hike by their ID.

# Note 
This might seem like a very long PR, but in reality most of it is the tests which require a long string for the requests.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -3,6 +3,7 @@ import java.io.FileInputStream
 import java.util.Properties
 
 plugins {
+    kotlin("plugin.serialization") version "1.8.10"
     alias(libs.plugins.androidApplication)
     alias(libs.plugins.jetbrainsKotlinAndroid)
     alias(libs.plugins.ktfmt)
@@ -146,6 +147,8 @@ dependencies {
     val composeBom = platform(libs.compose.bom)
 
     // Dependencies
+    implementation(libs.kotlinx.serialization.json.v163)
+    implementation(libs.kotlinx.serialization.json.v160)
     implementation(libs.androidx.navigation.compose)
     implementation(libs.androidx.core.ktx)
     implementation(libs.androidx.appcompat)
--- a/app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
+++ b/app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
@@ -21,8 +21,7 @@ private const val CONTENT_TYPE_HEADER = "Content-Type"
 private val JSON_MEDIA_TYPE = "application/json; charset=utf-8".toMediaType()
 
 /** Serializable data class for the response from the Elevation API */
-@Serializable
-data class ElevationResponse(val results: List<ElevationResult>)
+@Serializable data class ElevationResponse(val results: List<ElevationResult>)
 
 /** Serializable data class for the response from the Elevation API */
 @Serializable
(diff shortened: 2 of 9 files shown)


gradle/libs.versions.toml
```diff
@@ -1,8 +1,12 @@
 [versions]
 agp = "8.7.0"
 firebaseBom = "33.4.0"
+gson = "2.8.5"
+jetbrainsKotlinxSerializationJson = "1.6.3"
 kotlin =  "1.8.10"
 coreKtx = "1.13.1"
+kotlinxSerializationJson = "1.3.0"
+kotlinxSerializationJsonVersion = "1.6.0"
```
rBergCode, line 9: Question: Is there a reason not to used what @Wylwi used in #20 ? I'm talking about `android.util.JsonReader`.

If there is a good reason, I think it's fine. If not, maybe we would want to use the same "json serializer" for all API calls.

app/build.gradle.kts
```diff
@@ -146,6 +147,8 @@ dependencies {
     val composeBom = platform(libs.compose.bom)
 
     // Dependencies
+    implementation(libs.kotlinx.serialization.json.v163)
+    implementation(libs.kotlinx.serialization.json.v160)
```
rBergCode, line 151: Question: Why 2 versions ?

app/build.gradle.kts
```diff
@@ -3,6 +3,7 @@ import java.io.FileInputStream
 import java.util.Properties
 
 plugins {
+    kotlin("plugin.serialization") version "1.8.10"
```
rBergCode, line 6: Warning: Is this a hardcoded version in `build.gradle.kts` ? 👀
Also, you didn't add `jetbrainsKotlinxSerializationJson ` and `gson` package to the dependencies. Is there a reason ?

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+import okhttp3.RequestBody
+
+private const val BASE_URL = "https://api.open-elevation.com/api/v1/lookup"
+private const val LOCATION_KEY = "locations"
+private const val LATITUDE_KEY = "latitude"
+private const val LONGITUDE_KEY = "longitude"
+private const val ACCEPT_HEADER = "Accept"
+private const val APPLICATION_JSON = "application/json"
+private const val CONTENT_TYPE_HEADER = "Content-Type"
+private val JSON_MEDIA_TYPE = "application/json; charset=utf-8".toMediaType()
```
rBergCode, line 21: Love that!

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+@Serializable
+data class ElevationResult(val latitude: Double, val longitude: Double, val elevation: Double)
+
+/**
+ * A repository for the ElevationService. This class is responsible for making the network request
+ */
+class ElevationServiceRepository(private val client: OkHttpClient) : ElevationService {
+
+  // Cache for the elevation data, indexed by hike ID
+  private val cache = mutableMapOf<Long, List<Double>>()
```
rBergCode, line 36: Is the cache storing **elevation for all points of the hike** ?
I think we could discuss about how we want to use elevation. We might not need all elevation but 1 out of 4 for example.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+  override fun getElevation(
+      coordinates: List<LatLong>,
+      hikeID: Long,
+      onSuccess: (List<Double>) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    if (coordinates.isEmpty()) {
+      onFailure(Exception("No coordinates provided"))
+      return
+    }
```
rBergCode, line 56: I think we should return a `success response` with an empty list of doubles.
From my understanding, `getElevation` should return the list of **elevation** from the list of `LatLong`. If the list is empty, the return list should be empty.

Also, you forgot the docstring for this function.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+    // If the cache already has the elevation data for this hike, call onSuccess
+    for (mutableEntry in cache) {
+      if (mutableEntry.key == hikeID) {
+        onSuccess(mutableEntry.value)
+        return
+      }
+    }
```
rBergCode, line 63: This works fine but you may be reinventing the wheel here. You could use something like
```kotlin
if (cache.containsKey(hikeID)) {
    onSuccess(cache[hikeID])
    return
}
```

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+          buildJsonArray {
+            coordinates.forEach { coordinate ->
+              add(
+                  // Object with the location
+                  buildJsonObject {
+                    put(LATITUDE_KEY, coordinate.lat)
+                    put(LONGITUDE_KEY, coordinate.lon)
+                  })
+            }
+          })
```
rBergCode, line 78: Very clean!

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        onSuccess = {
+          result = it
+          latch.countDown()
+        },
+        onFailure = {
+          error = it
+          latch.countDown()
+        })
+
+    latch.await(10, TimeUnit.SECONDS) // Wait for the response or timeout
```
rBergCode, line 77: I understand that you have to wait for the API call to finish but this will really make the tests slow. I don't know if there is a better option. Have you done some research on this ?

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
```diff
@@ -0,0 +1,318 @@
...
+            .build()
+
+    // Mock the call
+    val call = mock(Call::class.java)
+    `when`(call.execute()).thenReturn(response)
+    `when`(client.newCall(any())).thenReturn(call)
+
+    elevationServiceRepository.getElevation(
+        latLong, 0, { fail("Expected failure but got success") }) { exception ->
+          assertEquals("Failed to get elevation", exception.message)
```
rBergCode, line 285: Nitpick: I don't think the message is relevant here. We just want to make sure the error is called. We don't want to change the test if we later change the exception message.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+    verify(client).newCall(any())
+  }
+
+  @Test
+  fun failsWithNoCoordinates() {
+    elevationServiceRepository.getElevation(
+        emptyList(), 0, { fail("Expected failure but got success") }) { exception ->
+          assertEquals("No coordinates provided", exception.message)
```
rBergCode, line 295: Same remark as the one on line 285

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    val call = mock(Call::class.java)
+    `when`(call.enqueue(any())).thenAnswer {
+      val callback = it.arguments[0] as Callback
+      callback.onFailure(call, IOException("Network error"))
+    }
+    `when`(client.newCall(any())).thenReturn(call)
+
+    elevationServiceRepository.getElevation(
+        latLong, 0, { fail("Expected failure but got success") }) { exception ->
+          assertEquals("Network error", exception.message)
```
rBergCode, line 313: Same remark as the one on line 285

rBergCode (changes requested):
Nice work ! Very nice and clean code. I have many questions about it but a big part of the work is good!

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    val body = RequestBody.create(JSON_MEDIA_TYPE, jsonBody.toString())
```
MaelImhof, line 81: Question: Sonar tells us that `create` is deprecated and we shouldn't use it. Is there an alternative?

mchac15: Same as for Rémy, I still need to do some changes specially in the build.gradle. Let me put it on draft for now

app/build.gradle.kts
mchac15: I won't change the code itself, let me look into the gradle issues

gradle/libs.versions.toml
mchac15: So:
1. The implementation is much easier to do and to understand the code.
2. I was having problems with the `JsonReader` since for some calls it would just return null when initialized: so this could totally be a problem from my side, but after some digging I found this solution much better.
Now my question to you: Is there any reason you would want me to use the `JsonReader`? If yes I could give it a try at redoing the implementation.

app/build.gradle.kts
mchac15: Sorry for this. I forgot to review the changes of the `build.gradle` all things were done as temporary I commited the fixes for these issues.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
mchac15: Yes this is totally correct. This is just a placeholder, since I really don't know how much data could be stored in elevation points. I will open a discussion in telegram.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
mchac15: Yes that is completeley right. Fixed

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
mchac15: Fixed.

mchac15: The reason I had it is because of a Sonar comment saying this code:
```kotlin
if (cache.containsKey(hikeID)) {
    onSuccess(cache[hikeID]!!)
    return
}
```
Was dangerous because of the null assertion.
Now I changed it to this:
```kotlin
if (cache.containsKey(hikeID)) {
    cache[hikeID]?.let { onSuccess(it) }
    return
}
```

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
```diff
@@ -0,0 +1,133 @@
...
+  ) : okhttp3.Callback {
+
+    override fun onFailure(call: okhttp3.Call, e: IOException) {
+      onFailure(e)
+    }
+
+    override fun onResponse(call: okhttp3.Call, response: okhttp3.Response) {
+      if (response.isSuccessful) {
+        val body = response.body?.string()
+        print(body)
```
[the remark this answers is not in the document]
mchac15: Sorry for that fixed.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
mchac15: You are right. Let me find a solution.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
mchac15: Solution:
```kotlin
  val jsonBodyString = jsonBody.toString()
  val body: RequestBody = jsonBodyString.toRequestBody(JSON_MEDIA_TYPE)

```
Sorry for this I checked Sonar and forgot to change it.

gradle/libs.versions.toml
rBergCode: No, it's good to me. We may want to use your solution for the OSM API too.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
mchac15: Fixed.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
mchac15: Fixed.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
mchac15: Fixed.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationServiceRepositoryTest.kt
mchac15: Fixed.

app/src/main/java/ch/hikemate/app/model/elevation/ElevationServiceRepository.kt
mchac15: From what was discussed in telegram: we will leave it like this for now and if there is performance/memory issues we'll change it later.

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
```diff
@@ -93,18 +84,12 @@ class ElevationIntegrationTest {
...
         onSuccess = {
           result = it
-          latch.countDown()
+          assertNotNull("Response should not be null", result)
+          assertEquals("There should be 3 elevation results", longCoordinates.size, result?.size)
         },
         onFailure = {
           error = it
-          latch.countDown()
+          assertTrue("Request failed: ${error?.message}", error == null)
```
rBergCode, line 92: Same remark as line 74

rBergCode (changes requested):
Just a quick remark on the tests. Thank you for your changes!

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
mchac15: Fixed

rBergCode (changes requested):
Sorry to bother but I may have given you a bad explanation.
I meant to **explictly** call a fail function in order to make the tests fails if `onFailure` is called. Like this:
```kotlin
onFailure = {
  fail("Should never be called")
}
``` 

In your case, if `onFailure` is called, your test will pass but shouldn't

app/src/test/java/ch/hikemate/app/model/elevation/ElevationIntegrationTest.kt
```diff
@@ -67,24 +65,17 @@ class ElevationIntegrationTest {
...
         onSuccess = {
           result = it
-          latch.countDown()
+          assertNotNull("Response should not be null", result)
+          assertEquals("There should be 3 elevation results", coordinates.size, result?.size)
         },
         onFailure = {
           error = it
-          latch.countDown()
+          assertTrue("Request failed: ${error?.message}", error == null)
```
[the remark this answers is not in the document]
mchac15: I removed it since it didn't make sense in this test. This is tested in the other test file.

mchac15:
> Sorry to bother but I may have given you a bad explanation. I meant to **explictly** call a fail function in order to make the tests fails if `onFailure` is called. Like this:
> 
> ```kotlin
> onFailure = {
>   fail("Should never be called")
> }
> ```
> 
> In your case, if `onFailure` is called, your test will pass but shouldn't

Fixed.

Issue #63, opened 2024-10-22
Create the SavedHikesViewModel and the SavedHikesRepository


MaelImhof:
Doesn't need to have a real implementation behind, just create the skeleton and hardcode some values, but think about the tests!
This will be needed for the OverviewScreen. See the User Story for required things to be included

MaelImhof:
`PlannedHikes` can be a confusing name since hikes that the user saves are not necessarily planned (planning is optional). I'm renaming this screen `SavedHikes`.

PR #66, opened 2024-10-22
feat(saved-hikes): add repository and view model

MaelImhof:
# Summary

This PR implements an initial draft of the repository and view model needed for the `Saved Hikes` screen.

The view model is working, but the repository will be implemented in a later PR (see #63, implementing the repository properly is not part of the task per se).

# Details

A saved hike consists of three things :
- The ID of the corresponding hike data, to retrieve the hike itself if needed and to uniquely identify saved hikes
- The human-readable name of the hike for display in the UI
- A scheduled date for the hike (meaning the user plans on going on this hike at that date. This is optional)

This data is represented in the data class `SavedHike`.

All changes have been made in the `ch.hikemate.app.model.route.saved` package.

Furthermore, a repository is in charge of storing, retrieving and modifying the saved hikes list. This repository offers operations to
- Retrieve the whole list of saved hikes
- Add a saved hike
- Remove a saved hike

I opted for a single function `getSavedHikes` instead of having pagination and retrieving the list part by part. I based this choice on the assumption that users probably won't save more than 1'000 hikes, which is perfectly fine for direct retrieval. I also assumed a hard maximum at 10'000 saved hikes, since Firebase can only return 10'000 documents per query. I'm not sure we should plan on users saving more than 10'000 hikes? Your call, I thought it was not worth the time and effort for such an edge case.

The above descriptions and decisions are also documented in the code directly, maybe even in a more detailed way.

I created a task on the SCRUM board to actually implement the repository and mentioned the hard limit of 10'000 documents, suggesting that a limit is put on how many hikes a user can save in Firebase itself using for example cloud functions. This is another task though.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebase.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebase.kt
@@ -1,0 +1,26 @@
+package ch.hikemate.app.model.route.saved
+
+import com.google.firebase.Timestamp
+
+class SavedHikesRepositoryFirebase : SavedHikesRepository {
+  override suspend fun loadSavedHikes(): List<SavedHike> {
+    return listOf(
+        SavedHike("1", "Hike 1", Timestamp.now()),
+        SavedHike("2", "Hike 2", Timestamp.now()),
+        SavedHike("3", "Hike 3", Timestamp.now()))
+  }
+
+  override suspend fun addSavedHike(hike: SavedHike) {
+    // A first implementation of the repository without any interaction with Firebase
+    // is implemented to allow the UI to be developed in parallel with the repository.
+    // There is nothing to do here except interacting with Firebase, hence this function
+    // is empty for now.
+  }
+
+  override suspend fun removeSavedHike(hike: SavedHike) {
+    // A first implementation of the repository without any interaction with Firebase
+    // is implemented to allow the UI to be developed in parallel with the repository.
+    // There is nothing to do here except interacting with Firebase, hence this function
+    // is empty for now.
+  }
+}
--- a/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
@@ -1,0 +1,84 @@
+package ch.hikemate.app.model.route.saved
+
+import android.util.Log
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class SavedHikesViewModel(
+    private val repository: SavedHikesRepository,
+    private val dispatcher: CoroutineDispatcher = Dispatchers.IO
+) : ViewModel() {
+
+  companion object {
+    /** Tag used for logging. */
+    private const val LOG_TAG = "SavedHikesViewModel"
+  }
+
+  private val _savedHikes = MutableStateFlow<List<SavedHike>>(emptyList())
+  /**
+   * The list of saved hikes for the user as a state flow. Observe this to get updates when the
+   * saved hikes change.
+   */
+  val savedHike: StateFlow<List<SavedHike>> = _savedHikes.asStateFlow()
+
+  /** Load saved hikes from the repository and update the [savedHike] state flow. */
+  fun loadSavedHikes() {
+    // Because loading saved hikes is also used by other methods, we extract it
+    viewModelScope.launch { loadSavedHikesAsync() }
+  }
+
+  /**
+   * Add the provided hike as a saved hike for the user.
+   *
+   * This function will update the [savedHike] state flow.
+   *
+   * @param hike The hike to add as a saved hike.
+   */
+  fun addSavedHike(hike: SavedHike) {
+    viewModelScope.launch(dispatcher) {
+      try {
+        repository.addSavedHike(hike)
+      } catch (e: Exception) {
+        Log.e(LOG_TAG, "Error adding saved hike", e)
+        return@launch
+      }
+      loadSavedHikesAsync()
+    }
+  }
+
+  /**
+   * Remove the provided hike from the saved hikes for the user.
+   *
+   * This function will update the [savedHike] state flow.
+   *
+   * @param hike The hike to remove from the saved hikes.
+   */
+  fun removeSavedHike(hike: SavedHike) {
+    viewModelScope.launch(dispatcher) {
+      try {
+        repository.removeSavedHike(hike)
+      } catch (e: Exception) {
+        Log.e(LOG_TAG, "Error removing saved hike", e)
+        return@launch
+      }
+      loadSavedHikesAsync()
+    }
+  }
+
+  private suspend fun loadSavedHikesAsync() {
+    withContext(dispatcher) {
+      try {
+        _savedHikes.value = repository.loadSavedHikes()
+      } catch (e: Exception) {
+        Log.e(LOG_TAG, "Error loading saved hikes", e)
+      }
+    }
+  }
+}
--- a/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
@@ -2,6 +2,7 @@ package ch.hikemate.app.model.route.saved
 
 import android.util.Log
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
 import ch.hikemate.app.R
 import kotlinx.coroutines.CoroutineDispatcher
@@ -18,6 +19,14 @@ class SavedHikesViewModel(
 ) : ViewModel() {
 
   companion object {
+    val Factory: ViewModelProvider.Factory =
+        object : ViewModelProvider.Factory {
+          @Suppress("UNCHECKED_CAST")
+          override fun <T : ViewModel> create(modelClass: Class<T>): T {
+            return SavedHikesViewModel(SavedHikesRepositoryFirebase()) as T
+          }
+        }
+
     /** Tag used for logging. */
     private const val LOG_TAG = "SavedHikesViewModel"
   }
--- a/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
@@ -3,6 +3,7 @@ package ch.hikemate.app.model.route.saved
 import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import ch.hikemate.app.R
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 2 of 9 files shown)


Wylwi:
> I opted for a single function getSavedHikes instead of having pagination and retrieving the list part by part. I based this choice on the assumption that users probably won't save more than 1'000 hikes, which is perfectly fine for direct retrieval. I also assumed a hard maximum at 10'000 saved hikes, since Firebase can only return 10'000 documents per query. I'm not sure we should plan on users saving more than 10'000 hikes? Your call, I thought it was not worth the time and effort for such an edge case.

I agree with this assumption. An argument in favor of it is the size of downloading such a data. The ID is a long (64 bits nowadays), the string will be 32 bits/character (UTF-8 can encode some usual character on 8 bits only, but let's get an upper bound) and the timestamp is, according to the doc, a long and an int, then 96 bits. Let's assume a 50 char string, then we have, for each hike, a 1760 bits (220 bytes) object. Retrieving 10'000 objects of this size equals retrieving around 2Mb of data, which is consequent for a real time retrieving function. So I agree on this limit

MaelImhof:
Those are precisely the computations I made, and the result I got. 2MB is indeed a significant amount of data, but :

- In all likelihood, it won't occur for most users, 10'000 / 365 > 27, meaning 10'000 hikes represents one hike per day for 27 years straight. Users might save hikes and never do them, but saving 10'000 hikes itself takes some time
- One complete fetch can be done at the app startup, and then the updates can be optimized while the app is running (Firebase listeners for example)

app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+    val Factory: ViewModelProvider.Factory =
+        object : ViewModelProvider.Factory {
+          @Suppress("UNCHECKED_CAST")
+          override fun <T : ViewModel> create(modelClass: Class<T>): T {
+            return SavedHikesViewModel(SavedHikesRepositoryFirebase()) as T
+          }
+        }
+
+    /** Tag used for logging. */
+    private const val LOG_TAG = "SavedHikesViewModel"
```
Wylwi, line 31: Question: Why is the tag for logging in the companion object? Couldn't be in the top level in private? I don't really know what is the cleanest place for it to be...

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.test.setMain
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Test
+
+class SavedHikesRepositoryFirebaseTest {
```
Wylwi, line 13: Same comment as before, this shouldn't be a RepositoryFirebaseTest, since there is no Firebase notion inside, even if this is still WIP

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  fun addSavedHikeDoesNothing() =
+      runTest(timeout = 5.seconds) {
+        // Add saved hike is a placeholder for now, it does nothing under the hood
+
+        // When
+        savedHikesRepositoryFirebase.addSavedHike(SavedHike("4", "Hike 4", null))
+
+        // Then
+        assert(true)
```
Wylwi, line 49: Nitpick: Couldn't you test something instead of asserting true? Like if the number of hikes is the same before and after the operation. Even if this isn't the expected behavior of a repository, it's the expected behavior of a dummy repository. (It then requires the file to be renamed to make sense and not confuse people)

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
```diff
@@ -0,0 +1,63 @@
...
+  @Test
+  fun removeSavedHikeDoesNothing() =
+      runTest(timeout = 5.seconds) {
+        // Remove saved hike is a placeholder for now, it does nothing under the hood
+
+        // When
+        savedHikesRepositoryFirebase.removeSavedHike(SavedHike("4", "Hike 4", null))
+
+        // Then
+        assert(true)
```
Wylwi, line 61: Nitpick: Couldn't you test something instead of asserting true? Like if the number of hikes is the same before and after the operation. Even if this isn't the expected behavior of a repository, it's the expected behavior of a dummy repository. (It then requires the file to be renamed to make sense and not confuse people)

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesViewModelTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    savedHikesRepository = mock(SavedHikesRepository::class.java)
+    savedHikesViewModel = SavedHikesViewModel(savedHikesRepository, UnconfinedTestDispatcher())
+  }
+
+  @Test
+  fun canBeCreatedAsFactory() {
+    val factory = SavedHikesViewModel.Factory
+    val viewModel = factory.create(SavedHikesViewModel::class.java)
+    assertNotNull(viewModel)
+  }
```
Wylwi, line 36: Great idea for a test!

Wylwi (commented):
Pretty good code for me! I even learned some things about testing corountines! I've got some semantics questions pending, so I'm not explicitly approving nor requesting changes, since it has to be discussed. 

Furthermore, I would recommend that you change your FirebaseRepository implementation to a DummyRepository implementation, since it is what it is, and that you enhance it a little bit, using a really simple implementation (using for example just a mutable list where you add and remove SavedRoutes). 
You could then improve your tests to make them a little bit more meaningful ! 

Otherwise, good work!

app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebase.kt
```diff
@@ -0,0 +1,26 @@
+package ch.hikemate.app.model.route.saved
+
+import com.google.firebase.Timestamp
+
+class SavedHikesRepositoryFirebase : SavedHikesRepository {
```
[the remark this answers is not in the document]
MaelImhof: Done in [f371674c](https://github.com/HikeMate/hikeMateApp/pull/66/commits/f371674cd55b28a97502642354fc44a7b7071f43).

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
MaelImhof: Done in [f371674c](https://github.com/HikeMate/hikeMateApp/pull/66/commits/f371674cd55b28a97502642354fc44a7b7071f43).

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
MaelImhof: Indeed it makes sense to have at least something going on under the hood. I added a mutable list and tests for the new (exciting) features of the dummy repository in [90e48b4c](https://github.com/HikeMate/hikeMateApp/pull/66/commits/90e48b4c0e6f6cbc37451c63160555d35e679da2).

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
MaelImhof: Indeed it makes sense to have at least something going on under the hood. I added a mutable list and tests for the new (exciting) features of the dummy repository in [90e48b4c](https://github.com/HikeMate/hikeMateApp/pull/66/commits/90e48b4c0e6f6cbc37451c63160555d35e679da2).

app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
MaelImhof: It makes sense for me that it should be private but other than that, I don't know where I should place it... I think I placed the tag for the map screen or some other thing I did in the companion object as well, but maybe we can change everywhere if it makes more sense.

MaelImhof: I found [this StackOverflow post](https://stackoverflow.com/a/45841698/) that states the constants should be defined in the companion object, but honestly it sounds like you do as you wish

MaelImhof: I have this [other StackOverflow post](https://stackoverflow.com/a/55054023) that states the contrary.

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
Wylwi: Lgtm! Thanks!

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryFirebaseTest.kt
Wylwi: Lgtm! Thanks!

app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
Wylwi: I think that creating a companion object can be done if there is something else inside, then it should not be a useless class generated by compiler. We could leave it like this for this class. I think we should not create a companion object only for constants. Thus, it's just a style convention...

MaelImhof: Yup, just found out top-level constants require the compiler to create another class, I think the companion object is clearer here (the constant is uniquely meant to be used in that class) and avoids creating a dummy class.

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryDummyTest.kt
```diff
@@ -40,24 +40,36 @@ class SavedHikesRepositoryFirebaseTest {
   @Test
   fun addSavedHikeDoesNothing() =
```
Wylwi, line 41: Really nitpicking: The name of the test is now incorrect 😅

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryDummyTest.kt
```diff
@@ -40,24 +40,36 @@ class SavedHikesRepositoryFirebaseTest {
...
+        savedHikesRepositoryDummy.addSavedHike(hike)
 
         // Then
-        assert(true)
+        assertEquals(4, savedHikesRepositoryDummy.loadSavedHikes().size)
+        assert(savedHikesRepositoryDummy.loadSavedHikes().contains(hike))
       }
 
   @Test
   fun removeSavedHikeDoesNothing() =
```
Wylwi, line 59: Really nitpicking: The name of the test is now incorrect 😅

Wylwi (commented):
Just really nitpicking on ridiculous details, just change the names and it should be good for merging!

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryDummyTest.kt
MaelImhof: Oh right thank you for noticing I forgot about those

app/src/main/java/ch/hikemate/app/model/route/saved/SavedHikesViewModel.kt
Wylwi: Looks like none of the stack overflow are relevant anymore. [Here](https://kotlinlang.org/docs/properties.html#compile-time-constants) is a link to the doc about it. From what I understood, the const val will be replaced as a macro do in C. But maybe, the companion object, if there is nothing else, will stay as a dummy class after it, whereas the top level declaration of the constants doesn't create such classes. 
Anyhow, since the companion object isn't empty here, it shouldn't have any consequences!

app/src/test/java/ch/hikemate/app/model/route/saved/SavedHikesRepositoryDummyTest.kt
MaelImhof: It's actually important imo, test names should represent what the test does, so thanks haha

Wylwi (approved):
Thanks for the updates!! Sorry for really nitpicking about some details! Your code is spotless and good! It looks good to me now!

PR #72, opened 2024-10-25
feat: fetch name and description from overpass API

Wylwi:
# Summary

- The objective was to have a name and a description from the OSM API to show to the user, fetched from the API

# Details

- Names are not always available. I've tried to implement ways to get a name, using alternative OSM tags and/or combination of tags to create decent name and/or tags, but this doesn't guarantee that a name can be shown AT ALL. 
- Some tags are more relevant than other, thus, there are priorities attributed to tags (For example, the `name` tag weights more than the `operator` tag when looking for a name.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
@@ -2,6 +2,7 @@ package ch.hikemate.app.model.route
 
 import android.util.JsonReader
 import android.util.Log
+import ch.hikemate.app.R
 import java.io.Reader
 import okhttp3.OkHttpClient
 import okhttp3.Request
@@ -229,8 +230,8 @@ class HikeRoutesRepositoryOverpass(val client: OkHttpClient) : HikeRoutesReposit
             nameEn != null -> nameEn
             otherName != null -> otherName
             from != null && to != null -> "$from - $to"
-            from != null -> from
-            to != null -> to
+            from != null -> "${R.string.hike_name_from_prefix} $from"
+            to != null -> "${R.string.hike_name_to_prefix} $to"
             else -> null
           }
 
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
@@ -201,7 +201,7 @@ class HikeRoutesRepositoryOverpass(val client: OkHttpClient) : HikeRoutesReposit
       // most relevant one
       var name: String? = null
       var description: String? = null
-      var nameFr: String? = null
+      var nameEn: String? = null
       var otherName: String? = null
       var from: String? = null
       var to: String? = null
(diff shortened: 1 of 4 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
```diff
@@ -168,10 +171,79 @@ class HikeRoutesRepositoryOverpass(val client: OkHttpClient) : HikeRoutesReposit
...
+      var otherName: String? = null
+      var from: String? = null
+      var to: String? = null
+      while (tagsReader.hasNext()) {
+        when (tagsReader.nextName()) {
+          // int_name is used for international names
+          "int_name" -> name = tagsReader.nextString()
+          // int_name has priority over name
+          "name" -> if (name == null) name = tagsReader.nextString() else tagsReader.skipValue()
+          "name:fr" -> nameFr = tagsReader.nextString()
```
rBergCode, line 214: Question: Why do you want a french name here ?

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
```diff
@@ -168,10 +171,79 @@ class HikeRoutesRepositoryOverpass(val client: OkHttpClient) : HikeRoutesReposit
...
+      }
+
+      val finalName =
+          when {
+            name != null -> name
+            nameFr != null -> nameFr
+            otherName != null -> otherName
+            from != null && to != null -> "$from - $to"
+            from != null -> from
+            to != null -> to
```
rBergCode, line 233: Suggestion: we may want to display something more helpful than "Genève" or "Lac Léman". I'm thinking os something like "Starting form Genève" or "Going to Lac Léman" would be more informative.

app/src/test/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpassTest.kt
```diff
@@ -474,7 +589,7 @@ class HikeRoutesRepositoryOverpassTest {
     }
 
     hikingRouteProviderRepositoryOverpass.getRoutes(
-        bounds, { routes -> assertEquals(simpleRoutes, routes) }) {
+        bounds, { routes -> assertEquals(alternativeRoutes, routes) }) {
```
rBergCode, line 592: Question: Why did you changed this ?

rBergCode (changes requested):
Nice code! Few questions and suggestions and you will be good to go!

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
Wylwi: Because it's not supposed to be french! I'll change that

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
Wylwi: Good idea! I'll add it

app/src/test/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpassTest.kt
Wylwi: Looks like I was tired and duplicated some code

rBergCode (approved):
LGTM!

PR #73, opened 2024-10-25
fix: change the name and logo of the app

Wylwi:
A sample name and a sample logo was still being used in the app, this small PR changes it with the correct ones and delete old files :)


MaelImhof (approved):
Such little changes, such a big and great impact ! It's good to open "HikeMate" on my phone instead of "MainActivity" or "SampleApp". Very nice !

We might want to have a vector for the logo eventually, so that it renders properly on all screen densities and resolutions, but for now I think those changes are strongly needed. Thanks for tackling this.

PR #75, opened 2024-10-27
fix(nav): localize button content description

MaelImhof:
# Summary

The content description of the button that opens the sidebar (main menu of the app) was hardcoded in the sidebar composable. A content description should

1. Be localized (translatable, i.e. defined in `strings.xml`
2. Describe an action (i.e. "Open sidebar menu" rather than "Sidebar")

I'm not 100% certain for 2., but 1. is certain. Hence this PR simply moves the definition of the sidebar button to `strings.xml`, where it can easily be translated.

It's a very simple PR, but I did not want to include it in the one I'm working on right now (Saved Hikes screen) as I felt it should be a PR of its own given its different nature.


mchac15 (commented):
LGTM

mchac15 (approved):
LGTM

PR #76, opened 2024-10-27
feat(nav): pass scaffold padding as content param

MaelImhof:
# Summary

The `SidebarNavigation` composable adds a burger menu button in the top left corner of the screen. For screens such as map, the fact that the button takes a bit of space hovering on the map is not a problem, the user can simply move the map (see attached image below).

However, on screens such as Saved Hikes, the burger menu button could hide the title of the screen or the top composable. This is not the desired behavior (see attached image below).

# Details

This PR proposes a quite simple solution to this. `SidebarNavigation` takes a `content` lambda as a parameter. Initially, this `content` lambda took no parameter. I changed it to take a parameter of type `PaddingValues`.

Existing code does not have to change. If there is no parameter in the provided lambda, Kotlin will assume the lambda wants to ignore the parameter, and it still compiles and works as expected. Tests all pass as well.

However, if one *wants* to use the padding, simply add `paddingValues ->` in the callback to take the parameter. It allows the map to keep its current behavior, and the Saved Hikes screen to take the menu button into account :

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/navigation/SideBarNavigation.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/navigation/SideBarNavigation.kt
@@ -49,7 +49,7 @@ fun SideBarNavigation(
     // The reason the content of the screen has to be passed as a lambda is because the drawer has
     // to be
     // integrated with the screen.
-    content: @Composable () -> Unit,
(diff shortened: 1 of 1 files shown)


Wylwi (approved):
Small changes, but that's what makes the difference! Good work! LGTM.

PR #79, opened 2024-10-28
fix(map): fix searching crashing when zoomed out

MaelImhof:
# Bug report

While testing our app manually, the coaches noticed that when one clicks `Search hikes here` when the map was zoomed out very far, the app crashed without any message.

# Bug investigation

When clicking `Search hikes here`, the app takes the bounding box of the OSMDroid map view (of type `BoundingBox`) and converts it to a `Bounds` instance (our own class). However, `Bounds` only accepts valid boundaries. The minimum latitude must be less than or equal to the maximum latitude. Same for longitude.

When the map was zoomed out very far, OSMDroid automatically repeated the world map on the left, right, top and bottom of the screen :

This lead the `BoundingBox` of the map view to be invalid in regards of the `Bounds` criteria, leading to an exception. This is because the map started at some minLatitude/minLongitude and ended at maxLatitude/maxLongitude, while minLatitude was possibly bigger than maxLatitude because it was on another map.

# Bug fix

I fixed the bug by disabling the repetition of the map when the user zoomed out a lot. I also adjusted the maximum zoom level so that the user does not see the blank background too much.

I also added a regression test to avoid re-introducing the bug in a future update. For this, I added optional parameters to the map screen, so that the tests can set the zoom levels from outside.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
@@ -30,39 +30,53 @@ class MapScreenTest : TestCase() {
 
   @get:Rule val composeTestRule = createComposeRule()
 
+  private fun setUpMap(
+      mapMinZoomLevel: Double = MapScreen.MAP_MIN_ZOOM,
+      mapInitialZoomLevel: Double = MapScreen.MAP_INITIAL_ZOOM
+  ) {
+    composeTestRule.setContent {
+      MapScreen(
+          hikingRoutesViewModel = listOfHikeRoutesViewModel,
+          navigationActions = navigationActions,
+          mapMinZoomLevel = mapMinZoomLevel,
+          mapInitialZoomLevel = mapInitialZoomLevel)
+    }
+  }
+
   @Before
   fun setUp() {
     navigationActions = mock(NavigationActions::class.java)
     hikesRepository = mock(HikeRoutesRepository::class.java)
     listOfHikeRoutesViewModel = ListOfHikeRoutesViewModel(hikesRepository)
-    composeTestRule.setContent {
-      MapScreen(
-          hikingRoutesViewModel = listOfHikeRoutesViewModel, navigationActions = navigationActions)
-    }
   }
 
   @Test
   fun mapIsDisplayed() {
+    setUpMap()
     composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsDisplayed()
   }
 
   @Test
   fun hikesListIsDisplayed() {
+    setUpMap()
     composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_HIKES_LIST).assertIsDisplayed()
   }
 
   @Test
   fun menuButtonIsDisplayed() {
+    setUpMap()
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
```diff
@@ -138,11 +157,29 @@ class MapScreenTest : TestCase() {
...
+
+  /**
+   * Regression test for the app crashing when searching for hikes while zoomed out too far.
+   *
+   * See https://github.com/HikeMate/hikeMateApp/issues/78 for more details.
+   */
+  @Test
+  fun zoomedOutMapDoesNotCrashWhenSearchingForHikes() {
+    // Initialize the map as zoomed out very far
+    setUpMap(mapMinZoomLevel = 1.0, mapInitialZoomLevel = 1.0)
```
Wylwi, line 177: Question: Just to clarify, you have a `setUpMap()` on every test because you need to alter the parameters here?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -87,6 +87,19 @@ object MapScreen {
...
+   */
+  const val MAP_MAX_ZOOM = 18.0
+
+  /**
+   * (Config) Minimum zoom level of the map. The zoom level is defined empirically to avoid the user
+   * zooming out too much and seeing too much of the blank background behind the map tiles while
+   * still being able to see a reasonable area of the world map.
+   */
+  const val MAP_MIN_ZOOM = 2.5
+
```
Wylwi, line 102: Nice use of constants!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -197,8 +214,14 @@ fun MapScreen(
...
-      controller.setZoom(MapScreen.MAP_INITIAL_ZOOM)
-      controller.setCenter(MapScreen.MAP_INITIAL_CENTER)
+      controller.setZoom(mapInitialZoomLevel)
+      controller.setCenter(mapInitialCenter)
+      // Limit the zoom to avoid the user zooming out or out too much
+      minZoomLevel = mapMinZoomLevel
+      maxZoomLevel = mapMaxZoomLevel
+      // Avoid repeating the map when the user reaches the edge or zooms out
+      isHorizontalMapRepetitionEnabled = false
+      isVerticalMapRepetitionEnabled = false
```
Wylwi, line 224: Nice fix!

Wylwi (approved):
Just one question, but I don't see any issues! LGTM!

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
MaelImhof: Yes, initially I could do that in the `setUp` function, but I need to change the zoom for this unique test so that it actually tests what is expected.

As a consequence, I had to move all the `setContent` calls to the tests themselves, otherwise I get an error `Error : cannot call setContent twice in a test`.

Wylwi (approved):
LGTM!

PR #89, opened 2024-10-29
fix(map): set user agent before OSMDroid map init

MaelImhof:
This PR introduces a quick fix for a minor bug that was causing Logcat errors when loading the map screen.

We use OSMDroid to display map tiles from OSM as a map view. OSMDroid requires to set an actual user agent for the tiles requests.

The configured user agent was indeed set to a valid value in our code, but it was set **after** the map view was created, meaning the first requests were being sent before the configuration was set, and an error could be seen in Logcat. It did not impact the functionality of the app too much, but is better fixed anyway.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
@@ -193,10 +193,6 @@ fun MapScreen(
     mapInitialCenter: GeoPoint = MapScreen.MAP_INITIAL_CENTER,
 ) {
   val context = LocalContext.current
-  // Avoid re-creating the MapView on every recomposition
-  val mapView = remember { MapView(context) }
-  // Keep track of whether a search for hikes is ongoing
-  var isSearching by remember { mutableStateOf(false) }
 
   // Only do the configuration on the first composition, not on every recomposition
   LaunchedEffect(Unit) {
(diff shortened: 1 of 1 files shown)


MaelImhof:
So the diff on this PR is pretty hard to read, here is a quick explanation of what I did.

The basic idea of the PR is that

```kotlin
userAgentValue = context.packageName
```

should come before

```kotlin
val mapView = remember { MapView(context) }
```

So that the user agent is set before initializing the map view.

The first line above has to occur only once at startup, not on every re-composition, hence it is placed in a `LaunchedEffect(Unit)` block. So I moved that block right after the definition of the `context` variable, which is needed in the block.

`mapView.apply` was also called inside of that `LaunchedEffect(Unit)` block, but it requires `mapView` to have been created. So I moved the `mapView.apply` block into the `remember` block of the map view's initialization.

# Code before changes

```kotlin
  val context = LocalContext.current
  // Avoid re-creating the MapView on every recomposition
  val mapView = remember { MapView(context) }
  // Keep track of whether a search for hikes is ongoing
  var isSearching by remember { mutableStateOf(false) }

  // Only do the configuration on the first composition, not on every recomposition
  LaunchedEffect(Unit) {
    Configuration.getInstance().apply {
      // Set user-agent to avoid rejected requests
      userAgentValue = context.packageName

      // Allow for faster loading of tiles. Default OSMDroid value is 2.
      tileDownloadThreads = 4

      // Maximum number of tiles that can be downloaded at once. Default is 40.
      tileDownloadMaxQueueSize = 40

      // Maximum number of bytes that can be used by the tile file system cache. Default is 600MB.
      tileFileSystemCacheMaxBytes = 600L * 1024L * 1024L
    }

    mapView.apply {
      // Set map's initial state
      controller.setZoom(mapInitialZoomLevel)
      controller.setCenter(mapInitialCenter)
      // Limit the zoom to avoid the user zooming out or out too much
      minZoomLevel = mapMinZoomLevel
      maxZoomLevel = mapMaxZoomLevel
      // Avoid repeating the map when the user reaches the edge or zooms out
      isHorizontalMapRepetitionEnabled = false
      isVerticalMapRepetitionEnabled = false
      // Disable built-in zoom controls since we have our own
      zoomController.setVisibility(CustomZoomButtonsController.Visibility.NEVER)
      // Enable touch-controls such as pinch to zoom
      setMultiTouchControls(true)
    }
  }
```

# Code after changes

```kotlin
  val context = LocalContext.current

  // Only do the configuration on the first composition, not on every recomposition
  LaunchedEffect(Unit) {
    Configuration.getInstance().apply {
      // Set user-agent to avoid rejected requests
      userAgentValue = context.packageName

      // Allow for faster loading of tiles. Default OSMDroid value is 2.
      tileDownloadThreads = 4

      // Maximum number of tiles that can be downloaded at once. Default is 40.
      tileDownloadMaxQueueSize = 40

      // Maximum number of bytes that can be used by the tile file system cache. Default is 600MB.
      tileFileSystemCacheMaxBytes = 600L * 1024L * 1024L
    }
  }

  // Avoid re-creating the MapView on every recomposition
  val mapView = remember {
    MapView(context).apply {
      // Set map's initial state
      controller.setZoom(mapInitialZoomLevel)
      controller.setCenter(mapInitialCenter)
      // Limit the zoom to avoid the user zooming out or out too much
      minZoomLevel = mapMinZoomLevel
      maxZoomLevel = mapMaxZoomLevel
      // Avoid repeating the map when the user reaches the edge or zooms out
      isHorizontalMapRepetitionEnabled = false
      isVerticalMapRepetitionEnabled = false
      // Disable built-in zoom controls since we have our own
      zoomController.setVisibility(CustomZoomButtonsController.Visibility.NEVER)
      // Enable touch-controls such as pinch to zoom
      setMultiTouchControls(true)
    }
  }
  // Keep track of whether a search for hikes is ongoing
  var isSearching by remember { mutableStateOf(false) }
```

rBergCode (approved):
LGTM!

PR #92, opened 2024-10-31
feat(map): display hike names instead of ids

MaelImhof:
# Summary

On the map screen, in the list of hikes below the map, instead of displaying each hike's ID, display their name instead. That is what this PR does.

Also added two tests to check that the name is displayed, or a default name if the name itself is not available.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
@@ -1,15 +1,20 @@
 package ch.hikemate.app.ui.map
 
+import androidx.compose.ui.test.assertCountEquals
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsNotDisplayed
+import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onAllNodesWithTag
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
+import androidx.test.platform.app.InstrumentationRegistry
+import ch.hikemate.app.R
 import ch.hikemate.app.model.route.Bounds
 import ch.hikemate.app.model.route.HikeRoute
 import ch.hikemate.app.model.route.HikeRoutesRepository
 import ch.hikemate.app.model.route.ListOfHikeRoutesViewModel
+import ch.hikemate.app.ui.components.TEST_TAG_HIKE_CARD_TITLE
 import ch.hikemate.app.ui.navigation.NavigationActions
 import ch.hikemate.app.ui.navigation.TEST_TAG_SIDEBAR_BUTTON
 import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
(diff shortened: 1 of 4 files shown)


app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -32,6 +32,8 @@ import androidx.compose.ui.unit.dp
 import ch.hikemate.app.R
 import ch.hikemate.app.ui.map.MapScreen
 
+const val TEST_TAG_HIKE_CARD_TITLE = "HikeCardTitle"
```
rBergCode, line 35: Warning: you should put this into an object so that it is not available project wide as `TEST_TAG_HIKE_CARD_TITLE` but as `HikeCard.TEST_TAG_HIKE_CARD_TITLE`

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
```diff
@@ -166,6 +171,64 @@ class MapScreenTest : TestCase() {
...
+    `when`(hikesRepository.getRoutes(any(), any(), any())).thenAnswer {
+      val onSuccess = it.getArgument<(List<HikeRoute>) -> Unit>(1)
+      onSuccess(listOf(hike))
+    }
+    val context = InstrumentationRegistry.getInstrumentation().targetContext
+
+    // When
+    setUpMap()
+    listOfHikeRoutesViewModel.setArea(BoundingBox(0.0, 0.0, 0.0, 0.0))
+    composeTestRule.waitUntil(timeoutMillis = 5000) {
```
rBergCode, line 216: Question: Is there another way to do that ? This would really make the tests longer in the long term

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
```diff
@@ -166,6 +171,64 @@ class MapScreenTest : TestCase() {
...
+    val hike = HikeRoute("1", Bounds(-1.0, -1.0, 1.0, 1.0), listOf(), hikeName, null)
+    `when`(hikesRepository.getRoutes(any(), any(), any())).thenAnswer {
+      val onSuccess = it.getArgument<(List<HikeRoute>) -> Unit>(1)
+      onSuccess(listOf(hike))
+    }
+
+    // When
+    setUpMap()
+    listOfHikeRoutesViewModel.setArea(BoundingBox(0.0, 0.0, 0.0, 0.0))
+    composeTestRule.waitUntil(timeoutMillis = 5000) {
```
rBergCode, line 187: Question: Is there another way to do that ? This would really make the tests longer in the long term

rBergCode (changes requested):
Good and simple PR. Well done!
Small remarks that need to be addressed and then you're good to go.

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
MaelImhof: I'm not sure what you mean with "that", are you pointing to the `waitUntil` clause? I'll look into another way to do this.

However there's probably only so much I can do, at the end of the day the test must wait until the interface is ready to be tested...

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
MaelImhof: Discussing this issue in the above conversation, closing this one.

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
rBergCode: Yes I meant the `waitUntil` (reviewing on a phone is not that good).
If you find a better way then go for it. If not, keep that

MaelImhof: I looked a bit into it

- `composeTestRule.awaitIdle()` and its synchronous equivalent `composeTestRule.waitForIdle()` are unreliable, tests do not pass for some reason
- `composeTestRule.waitUntil()` works but is not very clean I guess
- The solution I settled on is to use a coroutine dispatcher

I changed the tests to use an `UnconfinedTestDispatcher()`. This runs the view model's coroutines on the single main thread of the tests, meaning the test will naturally block and resume once the view model is done updating the UI (UI events update will block as well due to how `StateFlow` works).

However, this does not introduce any time gain, tests take the same time to execute (like, precisely the same time).

rBergCode: So, if I understand right, the test will naturally wait for the `setArea` to finish and then do the asserts ?
If so, how is there no time gain ? In your previous version, you waited for a fixed number of seconds which would eventually be longer than your new version.

rBergCode (approved):
LGTM now. Thanks!

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
MaelImhof: Actually I did not wait for a fixed number of seconds, I told `composeTestRule` to `waitUntil` (condition), and that condition was that there is a hike list item in the UI. The 5000ms delay was the timeout (time after which the test should fail if the condition is still not met).

rBergCode: Oh okay

PR #109, opened 2024-11-05
Feat: Implement the hike elevation graph component

Wylwi:
# Summary

- Implemented a hike elevation graph component
- Changed the HikeCard component to accept elevationData as a parameter

# Details

Depends on the nature of the PR :
- **Code changes :** describe the conceptual approach taken in the code and tell the reviewer what part of the changes does what, so that it's easier to read the code
- **Other changes :** Add details if relevant, or remove this section if not useful

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
@@ -1,0 +1,57 @@
+package ch.hikemate.app.ui.components
+
+import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import kotlin.math.ceil
+
+/**
+ * A composable function to display an elevation graph. The graph is averaged to ensure that the
+ * number of points is not too high, and thus that performance is not too impacted. The graph is
+ * drawn using a Canvas, which implies that it is not interactive.
+ *
+ * @param elevationData The list of elevation data
+ * @param maxNumberOfPoints The maximum number of points to display (default to 40)
+ * @param color The color of the graph (default to black)
+ * @param strokeWidth The width of the graph (default to 3f)
+ * @param modifier The modifier for the graph (default to fillMaxSize)
+ */
+@Composable
+fun ElevationGraph(
+    elevationData: List<Double>,
+    maxNumberOfPoints: Int = 40,
+    color: Color = Color.Black,
+    strokeWidth: Float = 3f,
+    modifier: Modifier = Modifier.fillMaxSize()
+) {
+  Canvas(modifier) {
+    // Averaging the elevation data to reduce the number of points
+    val averageSize = ceil(elevationData.size.toDouble() / maxNumberOfPoints).toInt()
+    val elevationDataAveraged = elevationData.chunked(averageSize) { chunk -> chunk.average() }
+
+    val height = size.height
+    val width = size.width
+
+    // Bunch of variables to scale the graph in height
+    val maxElevation = elevationDataAveraged.maxOrNull() ?: 0.0
+    val minElevation = elevationDataAveraged.minOrNull() ?: 0.0
+    val elevationRange = maxElevation - minElevation
+    val elevationStep = elevationRange / height
+
+    val widthStep = width / elevationDataAveraged.size
+
+    for (i in elevationDataAveraged.indices.take(elevationDataAveraged.size - 1)) {
+      // Scale the points to fit the whole width
+      val x = i * widthStep
+      // Scale the points to fit the whole height, but keeping the scale.
+      // The height is inverted because the origin of the canvas is at the top left corner
+      val y = (height - (elevationDataAveraged[i] - minElevation) / elevationStep).toFloat()
+      val nextX = (i + 1) * widthStep
+      val nextY = (height - (elevationDataAveraged[i + 1] - minElevation) / elevationStep).toFloat()
+      drawLine(color, Offset(x, y), Offset(nextX, nextY), strokeWidth = strokeWidth)
+    }
+  }
+}
--- a/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
@@ -2,7 +2,6 @@ package ch.hikemate.app.ui.components
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.Row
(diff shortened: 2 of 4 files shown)


app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -68,13 +95,15 @@ fun HikeCard(
...
-                Box(
+                ElevationGraph(
+                    elevationData = elevationData,
                     modifier =
                         Modifier.weight(1f)
                             .fillMaxHeight()
                             .background(
                                 MaterialTheme.colorScheme.primary.copy(alpha = 0.1f),
-                                RoundedCornerShape(4.dp)))
+                                RoundedCornerShape(4.dp)),
```
rBergCode, line 105: Warning: IMO, this should be in the `ElevationGraph` composable. This is global and should never change

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -0,0 +1,56 @@
...
+    for (i in elevationDataAveraged.indices.take(elevationDataAveraged.size - 1)) {
+      // Scale the points to fit the whole width
+      val x = i * widthStep
+      // Scale the points to fit the whole height, but keeping the scale.
+      // The height is inverted because the origin of the canvas is at the top left corner
+      val y = (height - (elevationDataAveraged[i] - minElevation) / elevationStep).toFloat()
+      val nextX = (i + 1) * widthStep
+      val nextY = (height - (elevationDataAveraged[i + 1] - minElevation) / elevationStep).toFloat()
+      drawLine(color, Offset(x, y), Offset(nextX, nextY), strokeWidth = strokeWidth)
+    }
```
rBergCode, line 54: Suggestion: This looks nice but not like the Figma. After some research I've seen a way to fill any polygones using the current clip. Can you look into that and check the performance afterward ?

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -0,0 +1,56 @@
...
+    elevationData: List<Double>,
+    modifier: Modifier = Modifier,
+    maxNumberOfPoints: Int = 40,
+    color: Color = Color.Black,
+    strokeWidth: Float = 3f,
+) {
+  Canvas(modifier) {
+    // Averaging the elevation data to reduce the number of points
+    val averageSize = ceil(elevationData.size.toDouble() / maxNumberOfPoints).toInt()
+    val elevationDataAveraged = elevationData.chunked(averageSize) { chunk -> chunk.average() }
```
rBergCode, line 32: Praise: Nice job!

rBergCode (changes requested):
Nice job! I would like you to look further about the graph in order to make it look more like the Figma.

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
Wylwi: Maybe not, we might want to change the appearance of the ElevationGraph (including the background) if we show it on other locations

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
Wylwi: True, this lead to performance issues, but I fixed it too

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -99,11 +100,15 @@ fun HikeCard(
                     elevationData = elevationData,
                     modifier =
                         Modifier.weight(1f)
+                            .zIndex(1.0f)
```
rBergCode, line 103: Question: Why did you add this ? After removing it, nothing changes in my side.

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -99,11 +100,15 @@ fun HikeCard(
...
                             .fillMaxHeight()
                             .background(
                                 MaterialTheme.colorScheme.primary.copy(alpha = 0.1f),
                                 RoundedCornerShape(4.dp)),
```
rBergCode, line 107: Suggestion: I tried removing the background and IMO it looks way better now that you filled the graphs. Could you remove it ?

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -14,7 +23,7 @@ import kotlin.math.ceil
  *
  * @param elevationData The list of elevation data
  * @param maxNumberOfPoints The maximum number of points to display (default to 40)
- * @param color The color of the graph (default to black)
+ * @param strokeColor The color of the graph (default to black)
  * @param strokeWidth The width of the graph (default to 3f)
  * @param modifier The modifier for the graph (default to fillMaxSize)
  */
```
rBergCode, line 29: Warning: You miss documentation for some parameters here

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -23,34 +32,68 @@ fun ElevationGraph(
...
+  val path = Path()
+  path.moveTo(scaledData.first().first, scaledData.first().second)
+
+  for ((x, y) in scaledData.drop(1)) {
+    path.lineTo(x, y)
+  }
+
+  val outlinePath = path.copy()
+  outlinePath.lineTo(width, height)
+  outlinePath.lineTo(0f, height)
```
rBergCode, line 87: Suggestion: I've tried to close the path so that the graph looks like in Figma. I think we should stick to the Figma design here. Here is my code to correct this:
```suggestion
  path.lineTo(width, height)
  path.lineTo(0f, height)
  path.close()
```

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -23,34 +32,68 @@ fun ElevationGraph(
...
+
+  Canvas(
+      modifier.onGloballyPositioned { coordinates ->
+        height = coordinates.size.height.toFloat()
+        width = coordinates.size.width.toFloat()
+      }) {
+        drawPath(path, strokeColor, style = Stroke(width = strokeWidth))
+        clipPath(outlinePath, clipOp = androidx.compose.ui.graphics.ClipOp.Intersect) {
```
rBergCode, line 95: Tied to last remark, should change to this:
```suggestion
        clipPath(path, clipOp = androidx.compose.ui.graphics.ClipOp.Intersect) {
```

rBergCode (changes requested):
Nice job of optimisation. Thanks!
I still have some remarks for you to correct. This should not take long and then you will be good to go!

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
Wylwi: Oh yeah, just forgot it during testing, thanks!

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
Wylwi: It isn't in the scope of this task to change the figma. As you stated in a previous comment, the code should reflect the figma, so we might want to change the figma first and then create a new task for it!

Wylwi: Yep, you're right!

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
Wylwi: Good catch thanks!

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
Wylwi: True, when removing the background, we can do that!

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
Wylwi: I reduced the stroke to make it less aggressive then

PR #118, opened 2024-11-07
feat(auth): implement the create account screen

rBergCode:
# Summary
Implemented the create account screen so that a user can create an account without Google.

Nothing more to say. The code is quite straightforward.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
@@ -1,6 +1,7 @@
 package ch.hikemate.app.ui.auth
 
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertTextContains
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
@@ -28,6 +29,7 @@ class CreateAccountScreenTest : TestCase() {
 
   private val TEST_NAME = "Test Name"
   private val TEST_EMAIL = "test@example.com"
+  private val TEST_WRONG_EMAIL = "test"
   private val TEST_PASSWORD = "password"
   private val TEST_DIFFERENT_PASSWORD = "different_password"
 
@@ -68,7 +70,14 @@ class CreateAccountScreenTest : TestCase() {
         .performTextInput(TEST_PASSWORD)
     composeTestRule
         .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
-        .performTextInput(TEST_PASSWORD)
+        .performTextInput(TEST_DIFFERENT_PASSWORD)
+
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_NAME_INPUT)
+        .assertTextContains(TEST_NAME)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT)
+        .assertTextContains(TEST_EMAIL)
   }
 
   @Test
@@ -89,6 +98,32 @@ class CreateAccountScreenTest : TestCase() {
 
     verify(authRepository, times(1))
         .createAccountWithEmailAndPassword(any(), any(), eq(TEST_EMAIL), eq(TEST_PASSWORD))
+  }
+
+  @Test
+  fun clickOnCreateAccountButtonWithEmptyFields() {
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
+
+    verify(authRepository, times(0)).createAccountWithEmailAndPassword(any(), any(), any(), any())
+  }
+
+  @Test
+  fun clickOnCreateAccountWithDummyEmail() {
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_NAME_INPUT)
+        .performTextInput(TEST_NAME)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT)
+        .performTextInput(TEST_WRONG_EMAIL)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput(TEST_PASSWORD)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
+        .performTextInput(TEST_PASSWORD)
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
+
+    verify(authRepository, times(0)).createAccountWithEmailAndPassword(any(), any(), any(), any())
   }
 
   @Test
--- a/app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
@@ -1,0 +1,113 @@
+package ch.hikemate.app.ui.auth
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextInput
+import ch.hikemate.app.model.authentication.AuthRepository
+import ch.hikemate.app.model.authentication.AuthViewModel
+import ch.hikemate.app.ui.components.BackButton
+import ch.hikemate.app.ui.navigation.NavigationActions
+import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.times
+import org.mockito.kotlin.any
+import org.mockito.kotlin.eq
+import org.mockito.kotlin.verify
+
+class CreateAccountScreenTest : TestCase() {
+  @get:Rule val composeTestRule = createComposeRule()
+
+  private lateinit var navigationActions: NavigationActions
+  private lateinit var authRepository: AuthRepository
+  private lateinit var authViewModel: AuthViewModel
+
+  private val TEST_NAME = "Test Name"
+  private val TEST_EMAIL = "test@example.com"
+  private val TEST_PASSWORD = "password"
+  private val TEST_DIFFERENT_PASSWORD = "different_password"
+
+  @Before
+  fun setUp() {
+    navigationActions = mock(NavigationActions::class.java)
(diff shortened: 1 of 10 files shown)


app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        .performTextInput(TEST_NAME)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT)
+        .performTextInput(TEST_EMAIL)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput(TEST_PASSWORD)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
+        .performTextInput(TEST_PASSWORD)
```
Wylwi, line 71: Question: Does this fail if it can't input text? Isn't there a cleaner way to do it? Like a "isEditable" or something like that? Didn't took time to look, just asking

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
+
+    verify(authRepository, times(1))
+        .createAccountWithEmailAndPassword(any(), any(), eq(TEST_EMAIL), eq(TEST_PASSWORD))
```
Wylwi, line 91: Remark: Not your task, but that doesn't make sense to have callbacks in the beginning of the method.... They should be at the end of the parameters list. Just so that we know it and think to put it as a new task

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        .performTextInput(TEST_PASSWORD)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
+        .performTextInput(TEST_DIFFERENT_PASSWORD)
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
+
+    verify(authRepository, times(0))
+        .createAccountWithEmailAndPassword(any(), any(), eq(TEST_EMAIL), eq(TEST_PASSWORD))
+  }
+}
```
Wylwi, line 113: Praise: Great edge case!!

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
Wylwi, line 1: Nitpick: Could you add a test for checking incorrect email? Like garbage data

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
```diff
@@ -50,10 +50,22 @@ class AuthViewModel(private val repository: AuthRepository) : ViewModel() {
...
+  fun createAccountWithEmailAndPassword(
+      email: String,
+      password: String,
+      onSuccess: () -> Unit,
+      onErrorAction: (Exception) -> Unit
+  ) {
+    if (email.isEmpty() || password.isEmpty()) {
+      onErrorAction(Exception("Email and password must not be empty"))
+      return
+    }
```
Wylwi, line 62: Nitpick: It's a great check to have here, but why don't you handle it **also** in the View? You already handle passwords not matching, and I think it would be cleaner to check for this issue too (with an additional check here!). It looks a bit "late" to me to wait for the ViewModel to throw an error like this, since it can be handled by the View too

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
```diff
@@ -0,0 +1,113 @@
...
+  fun everythingIsOnScreen() {
+    composeTestRule.onNodeWithTag(BackButton.BACK_BUTTON_TEST_TAG).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_TITLE).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_NAME_INPUT).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_PASSWORD_INPUT).assertIsDisplayed()
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
+        .assertIsDisplayed()
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).assertIsDisplayed()
```
Wylwi, line 55: Nitpick: You assigned a test tag to the Column (`Screen.CREATE_ACCOUNT`). Why isn't it checked here?

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -0,0 +1,150 @@
...
+                    password,
+                    onSuccess = {
+                      // Navigate to the map screen
+                      navigationActions.navigateTo(Route.MAP)
+                    },
+                    onErrorAction = {
+                      // Show an error message in a toast
+                      Toast.makeText(context, it.message, Toast.LENGTH_SHORT).show()
+                    })
+              }
```
Wylwi, line 147: Nitpick: For cleanliness, maybe just define an inner function called `onButtonPressed` that could be defined still in the CreateAccountScreen function, but higher up than in the Composable architecture, and call it here.

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -0,0 +1,150 @@
...
+            label = { Text(stringResource(R.string.create_account_repeat_password_label)) })
+
+        BigButton(
+            modifier = Modifier.fillMaxWidth().testTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON),
+            buttonType = ButtonType.PRIMARY,
+            label = stringResource(R.string.create_account_create_account_button),
+            onClick = {
+              if (password != confirmPassword) {
+                // Show an error message in a toast
+                Toast.makeText(context, mismatchErrorMessage, Toast.LENGTH_SHORT).show()
```
Wylwi, line 134: Nitpick: Since you do some checks on the password here, might be a good idea to just add an empty field check and a (quick) email check too. You can use this Regex for it:  `/^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$/`

Wylwi (changes requested):
Some minuscule changes needed, essentially nitpicking. Apart from that, looks like it's a pretty decent screen! Congrats for your work!

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
rBergCode: Haven't found anything like that. It doesn't look like `performTextInput` throw any error on such case. I will add some check about the content of the fields for this.
Nice catch btw!

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
rBergCode: This is mostly used for navigation (has to have a global tag for a whole screen). Do you think we should still check that the screen itself is displayed ?

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
rBergCode: Actually, as the passwords are obscured, I can do that check for these.
I keep it for name and email

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
rBergCode: I fixed it anyway

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
rBergCode: Nice idea. Thanks!

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
rBergCode: Nice idea. Thanks!

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
rBergCode: I did this as the firebase function will throw an error if given empty credentials.
I added a check in the view

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
rBergCode: Yep!

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
Wylwi: Okay 😂

app/src/androidTest/java/ch/hikemate/app/ui/auth/CreateAccountScreenTest.kt
Wylwi: If it's checked in another test, lgtm, since other tests tags implies it is displayed too

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -75,6 +78,33 @@ fun CreateAccountScreen(navigationActions: NavigationActions, authViewModel: Aut
...
+    when {
+      password != confirmPassword -> {
+        Toast.makeText(context, mismatchErrorMessage, Toast.LENGTH_SHORT).show()
+      }
+      email.isEmpty() || password.isEmpty() || name.isEmpty() -> {
+        Toast.makeText(context, fieldMustBeFilledErrorMessage, Toast.LENGTH_SHORT).show()
+      }
+      !email.matches("^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}\$".toRegex()) -> {
+        Toast.makeText(context, emailWrongFormatErrorMessage, Toast.LENGTH_SHORT).show()
+      }
```
Wylwi, line 91: Praise: Nice way to handle those!

Wylwi (approved):
LGTM! Thanks for the modifications!

PR #119, opened 2024-11-07
fix: add delay to instantaneous hike's search

Wylwi:
# Summary
- Searching for hikes were occurring instantly when retrieved from the cache. This commit introduces an artificial delay to simulate fetching the data correctly


rBergCode (approved):
LGTM! Nice fix

PR #121, opened 2024-11-07
fix(ui): fix status bar color

rBergCode:
# Summary
The color of the status bar was purple. I changed it to match the background color (white or black depending if dark mode is activated or not)


Wylwi (approved):
LGTM

PR #122, opened 2024-11-07
fix: allow new google accounts to sign up

Wylwi:
# Summary
By default, Google Sign in limits the sign in to accounts already registered once. This PR fix it


MaelImhof (approved):
That fixes the bug for me, thank you very much !

PR #126, opened 2024-11-09
fix(auth): fix go to map directly when signed in

rBergCode:
# Summary
Fixed a bug where when a user was logged in, the sign in screen would appear before directing him to the map screen.

# Details
I also cleaned up a little bit the `SignInScreen` as it was receiving a useless parameter that was never used.
Also, a lot of changes in `MainActivity` are due to formatting because I made a line longer.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/auth/SignInScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/auth/SignInScreenTest.kt
@@ -77,8 +77,7 @@ class SignInScreenTest : TestCase() {
 
   @Test
   fun whenUserIsSignedIn_navigatesToMap() {
-    val mockUser = mockk<FirebaseUser>(relaxed = true)
-    mockUserStateFlow.value = mockUser
+    every { mockAuthViewModel.isUserLoggedIn() } returns true
 
     setupSignInScreen()
 
--- a/app/src/main/java/ch/hikemate/app/MainActivity.kt
+++ b/app/src/main/java/ch/hikemate/app/MainActivity.kt
@@ -54,45 +54,52 @@ fun HikeMateApp() {
 
   val authViewModel = AuthViewModel(FirebaseAuthRepository())
 
-  NavHost(navController = navController, startDestination = TopLevelDestinations.AUTH.route) {
-    navigation(
-        startDestination = Screen.AUTH,
-        route = Route.AUTH,
-    ) {
-      composable(Screen.AUTH) { SignInScreen(navigationActions, authViewModel) }
-      composable(Screen.SIGN_IN_WITH_EMAIL) {
-        SignInWithEmailScreen(navigationActions, authViewModel)
-      }
-      composable(Screen.CREATE_ACCOUNT) { CreateAccountScreen(navigationActions, authViewModel) }
-    }
+  val isUserLoggedIn = authViewModel.isUserLoggedIn()
 
-    navigation(
-        startDestination = Screen.SAVED_HIKES,
-        route = Route.SAVED_HIKES,
-    ) {
-      composable(Screen.SAVED_HIKES) { SavedHikesScreen(navigationActions = navigationActions) }
-    }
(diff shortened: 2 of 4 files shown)


app/src/androidTest/java/ch/hikemate/app/ui/auth/SignInScreenTest.kt
```diff
@@ -77,8 +77,7 @@ class SignInScreenTest : TestCase() {
 
   @Test
   fun whenUserIsSignedIn_navigatesToMap() {
-    val mockUser = mockk<FirebaseUser>(relaxed = true)
-    mockUserStateFlow.value = mockUser
+    every { mockAuthViewModel.isUserLoggedIn() } returns true
```
davidholzwarth, line 80: Praise: LGTM, the change makes sense given the updated mockAuthViewModel. :)

app/src/main/java/ch/hikemate/app/MainActivity.kt
```diff
@@ -54,45 +54,52 @@ fun HikeMateApp() {
...
-        startDestination = Screen.AUTH,
-        route = Route.AUTH,
-    ) {
-      composable(Screen.AUTH) { SignInScreen(navigationActions, authViewModel) }
-      composable(Screen.SIGN_IN_WITH_EMAIL) {
-        SignInWithEmailScreen(navigationActions, authViewModel)
-      }
-      composable(Screen.CREATE_ACCOUNT) { CreateAccountScreen(navigationActions, authViewModel) }
-    }
+  val isUserLoggedIn = authViewModel.isUserLoggedIn()
```
davidholzwarth, line 57: Question: I am confused about what is going on here, did the code actually change? I think its just GitHub convoluting the changes to look more than they actually are, right?

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
```diff
@@ -22,6 +22,11 @@ class AuthViewModel(private val repository: AuthRepository) : ViewModel() {
   val currentUser: StateFlow<FirebaseUser?>
     get() = _currentUser
 
+  /** Checks if the user is currently logged in. */
+  fun isUserLoggedIn(): Boolean {
+    return _currentUser.value != null
+  }
```
davidholzwarth, line 28: Praise: LGTM! Useful function too, especially for testing. :)

app/src/main/java/ch/hikemate/app/ui/auth/SignInScreen.kt
```diff
@@ -88,8 +83,8 @@ fun SignInScreen(
           }
 
   // If the user is already signed in, navigate to the map screen
-  LaunchedEffect(currUser) {
-    if (currUser != null) {
+  LaunchedEffect(authViewModel.isUserLoggedIn()) {
+    if (authViewModel.isUserLoggedIn()) {
```
davidholzwarth, line 87: Praise: Great, using only the VMs currUser variable is soo much cleaner, thanks for the fix!

davidholzwarth (commented):
Thanks Remy. If you could just quickly explain what exactly actually changed in the MainActivity (bc thats whats actually important in the PR, right?) then you're good to merge.

app/src/main/java/ch/hikemate/app/MainActivity.kt
rBergCode: Yes, as I said in the PR description, the whole code is in change because it changes the formatting. I only changed the code to use `isUserLoggesIn` to decide if we display `AUTH` or `MAP`. The difference is that we now do that in `MainActivity` instead of `SignInScreen` so that the Sign in screen will no longer show up.

davidholzwarth: I see, thanks for clarifying.

davidholzwarth (approved):
Thanks for clearing everything up.

PR #128, opened 2024-11-09
fix(test): fix nav and end2end tests

rBergCode:
# Summary
Fixed the navigation and end2end tests using the new sign in with email feature.

# Details
For this I used `Thread.sleep(1000)` in order to make sure the user is signed in. If you have a better idea I would love to here about it. I made sure to make this check as least as possible (only 1 per test) in order to not make the CI run for too long.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
@@ -24,6 +24,7 @@ import ch.hikemate.app.ui.navigation.TEST_TAG_SIDEBAR_BUTTON
 import ch.hikemate.app.ui.navigation.TopLevelDestinations
 import ch.hikemate.app.ui.saved.TEST_TAG_SAVED_HIKES_SECTION_CONTAINER
 import com.google.firebase.FirebaseApp
+import com.google.firebase.auth.EmailAuthProvider
 import com.google.firebase.auth.FirebaseAuth
 import java.util.UUID
 import junit.framework.TestCase
@@ -54,6 +55,8 @@ class EndToEndTest : TestCase() {
   @After
   fun deleteUser() {
     // Sign out after deleting for sanity check and un-reliability
+    val credential = EmailAuthProvider.getCredential(email, password)
+    auth.currentUser?.reauthenticate(credential)
     auth.currentUser?.delete()
     auth.signOut()
   }
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
@@ -1,15 +1,21 @@
 package ch.hikemate.app.endtoend
 
+import android.content.Context
 import androidx.compose.ui.test.assertHasClickAction
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextInput
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.hikemate.app.MainActivity
 import ch.hikemate.app.ui.auth.SignInScreen
+import ch.hikemate.app.ui.auth.SignInWithEmailScreen
 import ch.hikemate.app.ui.map.MapScreen.TEST_TAG_MAP
 import ch.hikemate.app.ui.navigation.LIST_TOP_LEVEL_DESTINATIONS
+import ch.hikemate.app.ui.navigation.Screen
 import ch.hikemate.app.ui.navigation.Screen.PROFILE
 import ch.hikemate.app.ui.navigation.TEST_TAG_DRAWER_CLOSE_BUTTON
 import ch.hikemate.app.ui.navigation.TEST_TAG_DRAWER_CONTENT
@@ -17,24 +23,55 @@ import ch.hikemate.app.ui.navigation.TEST_TAG_DRAWER_ITEM_PREFIX
 import ch.hikemate.app.ui.navigation.TEST_TAG_SIDEBAR_BUTTON
 import ch.hikemate.app.ui.navigation.TopLevelDestinations
 import ch.hikemate.app.ui.saved.TEST_TAG_SAVED_HIKES_SECTION_CONTAINER
+import com.google.firebase.FirebaseApp
+import com.google.firebase.auth.FirebaseAuth
+import junit.framework.TestCase
+import org.junit.After
+import org.junit.Before
 import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
 
-// @RunWith(AndroidJUnit4::class)
-// TODO: Change it to "EndToEndTest: TestCase()"
-class EndToEndTest {
+@RunWith(AndroidJUnit4::class)
+class EndToEndTest: TestCase() {
   @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
+  private val auth = FirebaseAuth.getInstance()
+  private val email = "end2endTest@gmail.com"
+  private val password = "password"
 
-  // @Test
+  @Before
+  fun setupFirebase() {
+    val context = ApplicationProvider.getApplicationContext<Context>()
+    FirebaseApp.initializeApp(context)
+
+    auth.createUserWithEmailAndPassword(email, password)
+    auth.signOut()
+  }
+
+  @After
+  fun deleteUser() {
+    auth.currentUser?.delete()
+    auth.signOut()
+  }
+
+  @Test
   fun test() {
     // Check that we are on the login screen
     composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_EMAIL).assertIsDisplayed()
     composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).assertIsDisplayed()
 
-    // Click on the login button
-    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).performClick()
+    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_EMAIL).performClick()
+    composeTestRule.onNodeWithTag(Screen.SIGN_IN_WITH_EMAIL).assertIsDisplayed()
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_EMAIL_INPUT)
+        .performTextInput(email)
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput(password)
+    composeTestRule.onNodeWithTag(SignInWithEmailScreen.TEST_TAG_SIGN_IN_BUTTON).performClick()
 
-    // Check that we are not on the login screen anymore
-    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).assertIsNotDisplayed()
+    // Wait for the map to load
+    Thread.sleep(1000)
 
     // Check that we are on the map
     composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsDisplayed()
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
@@ -53,6 +53,7 @@ class EndToEndTest : TestCase() {
 
   @After
   fun deleteUser() {
+    // Sign out after deleting for sanity check and un-reliability
     auth.currentUser?.delete()
     auth.signOut()
   }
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
+  private val password = "password"
 
-  // @Test
+  @Before
+  fun setupFirebase() {
+    val context = ApplicationProvider.getApplicationContext<Context>()
+    FirebaseApp.initializeApp(context)
+
+    auth.createUserWithEmailAndPassword(email, password)
+    auth.signOut()
```
Wylwi, line 48: Question: Why is signOut necessary here? Does createUserWithEmailAndPassword automatically logs in the user?

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
+  }
+
+  @After
+  fun deleteUser() {
+    auth.currentUser?.delete()
+    auth.signOut()
```
Wylwi, line 54: Question: Is signOut necessary after deleting the user?

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
+    val context = ApplicationProvider.getApplicationContext<Context>()
+    FirebaseApp.initializeApp(context)
+
+    auth.createUserWithEmailAndPassword(email, password)
+    auth.signOut()
+  }
+
+  @After
+  fun deleteUser() {
+    auth.currentUser?.delete()
```
Wylwi, line 53: Nitpick: I know that delete can fail if the user haven't (re)authenticated recently. Since this is supposed to be cleanup, could you maybe add a relogin to ensure that the account will be deleted correctly? Or at least an error handling method that will retry delete correctly if there is any kind of issue?

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
 
-// @RunWith(AndroidJUnit4::class)
-// TODO: Change it to "EndToEndTest: TestCase()"
-class EndToEndTest {
+@RunWith(AndroidJUnit4::class)
+class EndToEndTest : TestCase() {
   @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
+  private val auth = FirebaseAuth.getInstance()
+  private val email = "end2endTest@gmail.com"
+  private val password = "password"
```
Wylwi, line 40: Nitpick: Maybe add a UUID for the email, such that this test can be run by two people at the same time? If one test delete the user while another one is going through the screen, this will cause issues....

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
+        .performTextInput(email)
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput(password)
+    composeTestRule.onNodeWithTag(SignInWithEmailScreen.TEST_TAG_SIGN_IN_BUTTON).performClick()
 
-    // Check that we are not on the login screen anymore
-    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).assertIsNotDisplayed()
+    // Wait for the map to load
+    Thread.sleep(1000)
```
Wylwi, line 74: Question: Have you tried if `composeTestRule.waitForIdle()` isn't enough? It would be better that a `Thread.sleep` if it works

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -1,40 +1,77 @@
...
+  @Test
   fun test() {
     // Check that we are on the login screen
     composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_EMAIL).assertIsDisplayed()
     composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).assertIsDisplayed()
 
-    // Click on the login button
-    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).performClick()
+    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_EMAIL).performClick()
+    composeTestRule.onNodeWithTag(Screen.SIGN_IN_WITH_EMAIL).assertIsDisplayed()
```
Wylwi, line 64: Praise: Nice check!!

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
+import org.junit.runner.RunWith
 
-// @RunWith(AndroidJUnit4::class)
+@RunWith(AndroidJUnit4::class)
 class HikeMateAppNavigationTest {
   // Set up the Compose test rule
   @get:Rule val composeTestRule = createComposeRule()
+  private val auth = FirebaseAuth.getInstance()
+  private val email = "navTesting@gmail.com"
+  private val password = "password"
```
Wylwi, line 33: Nitpick: Same as above, recommend usage of a UUID

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
+    auth.signOut()
+  }
 
-    composeTestRule.onNodeWithTag(Screen.AUTH).assertIsDisplayed()
+  private fun setupUser() {
+    val context = ApplicationProvider.getApplicationContext<Context>()
+    FirebaseApp.initializeApp(context)
+
+    auth.createUserWithEmailAndPassword(email, password)
+    auth.signOut()
```
Wylwi, line 46: Same as above about signOut after createUserWithEmailAndPassword

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
   }
 
-  // @Test
-  fun testNavigationToMapScreen() {
-    composeTestRule.setContent { HikeMateApp() }
+  private fun deleteUser() {
+    auth.currentUser?.delete()
+    auth.signOut()
```
Wylwi, line 51: Same as above about signOut after delete

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_GO_TO_SIGN_UP_BUTTON)
+        .performClick()
+    composeTestRule.onNodeWithTag(Screen.CREATE_ACCOUNT).assertIsDisplayed()
+
+    // Navigate back to the sign in screen
+    composeTestRule.onNodeWithTag(BackButton.BACK_BUTTON_TEST_TAG).performClick()
+    composeTestRule.onNodeWithTag(Screen.SIGN_IN_WITH_EMAIL).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(BackButton.BACK_BUTTON_TEST_TAG).performClick()
+    composeTestRule.onNodeWithTag(Screen.AUTH).assertIsDisplayed()
```
Wylwi, line 70: Praise: Nice test! It really tests the navigation and checks that the back button goes only back one screen!

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_EMAIL_INPUT)
+        .performTextInput(email)
+    composeTestRule
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput(password)
+    composeTestRule.onNodeWithTag(SignInWithEmailScreen.TEST_TAG_SIGN_IN_BUTTON).performClick()
+
+    // Wait for the map to load
+    Thread.sleep(1000)
```
Wylwi, line 95: Same as above about `waitForIdle`

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -1,67 +1,112 @@
...
-    composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_GOOGLE).performClick()
-    composeTestRule.onNodeWithTag(Screen.MAP).assertIsDisplayed()
-    // Open the sidebar
+    // Go to profile screen
     composeTestRule.onNodeWithTag(TEST_TAG_SIDEBAR_BUTTON).performClick()
     composeTestRule.onNodeWithTag(TEST_TAG_DRAWER_ITEM_PREFIX + Route.PROFILE).performClick()
     composeTestRule.onNodeWithTag(Screen.PROFILE).assertIsDisplayed()
+
+    // Delete the user
+    deleteUser()
```
Wylwi, line 110: Warning: If the test fails, the `deleteUser` is not called. You must find a way to call it even when an assert fails

Wylwi (changes requested):
Overall, pretty good changes, with a finally working EndToEnd tests. Only small changes to do about when cleanup on test fail and about tests running in parallel. Should not be too long!!

Also, it's out of the scope of this PR, but be aware of the [following issue](https://github.com/orgs/HikeMate/projects/1/views/2?pane=issue&itemId=86669022), that I created in the Backlog to address it at some point.

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
rBergCode: Yes it does!

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
rBergCode: Deleting the user should signing him out. However, when testing, it could be unreliable so I added this sanity check here.

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
rBergCode: Good idea!

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
rBergCode: Same as above, yes

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
rBergCode: Same as above

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
rBergCode: Same as above

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
rBergCode: I just tried and this doesn't seems to work. I think 1 second delay is fair for this kind of test (we won't have much of them)

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
rBergCode: Yes you're right!

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
Wylwi: Could you add a small comment about it? So that if people don't read this PR, they know why it has been done this way

rBergCode: Sure

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
Wylwi: I agree, but we can't make assumptions about the cleanup of test users. Don't you have a way to check if delete accomplished correctly?

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
Wylwi: Same as above too :)

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
Wylwi: Okay, thanks for the try!

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
rBergCode: There are 2 exceptions that could be thrown:
The first one is when the user doesn't exist (which is not a problem in our case)
The second one is if the user is not re-authenticate -> I forced it to be

Wylwi (approved):
Thanks for the changes! Your code is great and we now have a working E2E test!

PR #135, opened 2024-11-12
fix(auth): fix sign in with google

rBergCode:
# Summary
Fixed signing in with Google which would not lead to the map screen.


Wylwi (approved):
LGTM. Quite a subtle issue!

PR #147, opened 2024-11-13
fix: remove inappropriate chars in hike names

Wylwi:
# Summary
The Overpass API can return names containing fixme, question mark and other inappropriate chars for a production app. This PR removes them from user's view

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
@@ -23,7 +23,7 @@ private const val MAX_DISTANCE_BETWEEN_OPENED_PATH = 10 // meters
  *
  * @see <a href="https://dev.overpass-api.de/overpass-doc/">Overpass API documentation</a>
  */
-class HikeRoutesRepositoryOverpass(val client: OkHttpClient) : HikeRoutesRepository {
+class HikeRoutesRepositoryOverpass(private val client: OkHttpClient) : HikeRoutesRepository {
   private val cachedHikeRoutes = mutableMapOf<Bounds, List<HikeRoute>>()
 
   /** @return The size of the cache. */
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
```diff
@@ -365,9 +365,7 @@ class HikeRoutesRepositoryOverpass(private val client: OkHttpClient) : HikeRoute
           // int_name has priority over name
           "name" -> if (name == null) name = tagsReader.nextString() else tagsReader.skipValue()
           "name:en" -> nameEn = tagsReader.nextString()
-          "osmc:name",
-          "operator",
-          "symbol" ->
+          "osmc:name" ->
```
rBergCode, line 368: Question: Why do you stop using `operator` or `symbol` ? Is it because they always are not interesting ?

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
```diff
@@ -389,10 +387,14 @@ class HikeRoutesRepositoryOverpass(private val client: OkHttpClient) : HikeRoute
...
       // We also assert that the name is not the same as the description
-      if (description == null && from != null && to != null && finalName != "$from - $to") {
+      if (description == null && from != null && to != null && selectedName != "$from - $to") {
         description = "$from - $to"
       }
 
+      // Remove fix me from the name
+      val finalName =
+          selectedName?.replace(Regex("""(\? - fixme|fixme - | - fixme| fixme|fixme |[?])"""), "")
+
```
rBergCode, line 397: Question: Nice regex! Have you done some research about other "unwanted names" ?

rBergCode (commented):
Nice job! Please answer my questions and you will be good to go!

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
Wylwi: Exactly! I just saw that doing so leads to having multiple "Yellow diamond" (From the name of the symbol on the hike) hikes, which is (I think), worse that "Unnamed hike", especially when there is like 5 or 6 "Yellow diamond" hikes following

app/src/main/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpass.kt
Wylwi: Since it's defined by user themselves, the only research I've been able to do is by looking around on the maps for names that are incorrects

rBergCode (approved):
Thanks for your answers!

Issue #146, opened 2024-11-13
Connect Hike Details screen to the actual saved hikes repository


mchac15:
This requires the Saved Hikes repository to be actually implemented using Firebase.

Once it is implemented, the hike details screen implementation should be updated to use the `isHikeSaved` and `getSavedHike` methods where needed.

**Definition of Done**

- [x] The Hike Details screen displays whether the hike is actually saved, not a dummy value
- [x] The Hike Details screen displays the actual name and date of the hike, not a dummy value

PR #148, opened 2024-11-13
fix: firebase can't deserialize saved hikes

Wylwi:
Firebase throws an error when the object to deserialize don't have an empty constructor. This simple PR add a default argument for `SavedHike` and `UserSavedHike`

**Since this PR is needed urgently, the reviewer is allowed to merge it himself after approving**


mchac15 (approved):
The bug was reported after the `HikeDetailsScreen.kt` in #146.
This issue is now solved.

PR #150, opened 2024-11-14
Feat/more hike details for list of hike routes view model

davidholzwarth:
# Summary

Add a DetailedHikeRoute data class which contains more information about a hikeRoute:

- Distance
- Elevation gain
- Estimated time
- Difficulty

This PR also immediately integrates this functionality into the HikeDetails screen.

# Details

In order to get the code more modular and keep the ListOfHikeRoutesViewModel cleaner, I have opted to separate this functionality from the viewModel, and put the entire functionality into a companion object function `create` for this data class. When calling `DetailedHikeRoute.create(HikeRoute)`, the corresponding DetailedHikeRoute is returned.


davidholzwarth:
@Reviewer, this version of the code still has some logs from development and might not be completely clean, I have not yet done this final aesthetic check, however I pushed this PR now so that you can do the PR and we don't get into time constraints this afternoon before the deadline. Just so you know, incase you find any ugly things 😅

app/src/test/java/ch/hikemate/app/model/route/DetailedHikeRouteTest.kt
rBergCode, line 1: Nice tests! 
Did you do the computation of the values by hand ? If not, I think you probably use your implementation to compute them (in which case, those tests don't make any sense.)

If it is the case, I would propose to remove the test about total distance, estimated time and elevation gain and keep the one about difficulty (which is easy to test)

Also, I think you should do some mocking for the elevation API. You can hardcode the elevation (which would make the testing easier also)

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -49,6 +52,8 @@ class HikeDetailScreenTest {
           description =
               "A scenic trail with breathtaking views of the Matterhorn and surrounding glaciers.")
 
+  private val detailedRoute = DetailedHikeRoute.create(route)
```
rBergCode, line 55: Warning: I think doing API calls for testing is a bad practice. You should mock it and hardcode the returned elevations.

rBergCode (changes requested):
Very nice job! You did that very quickly and well.
I have some comments that you should change before it's ready.

rBergCode (approved):
Thanks for the changes! LGTM!

PR #157, opened 2024-11-14
Fix/sign in with google

rBergCode:
# Summary
When creating an account or signing in with google, if the profile was already created the app would not move the map screen.

# Details
While fixing this, I also remove the `addProfile` function as it was useless in our case

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/profile/ProfileRepositoryFirestore.kt
+++ b/app/src/main/java/ch/hikemate/app/model/profile/ProfileRepositoryFirestore.kt
@@ -48,7 +48,19 @@ class ProfileRepositoryFirestore(private val db: FirebaseFirestore) : ProfileRep
     val displayName = fireUser.displayName ?: context.getString(R.string.default_display_name)
     val email = fireUser.email ?: context.getString(R.string.default_email)
     val profile = Profile(fireUser.uid, displayName, email, HikingLevel.BEGINNER, Timestamp.now())
-    addProfile(profile, onSuccess = { onSuccess(profile) }, onFailure = onFailure)
+
+    profileExists(
+        fireUser.uid,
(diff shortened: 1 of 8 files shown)


app/src/main/java/ch/hikemate/app/model/profile/ProfileRepositoryFirestore.kt
```diff
@@ -39,17 +38,29 @@ class ProfileRepositoryFirestore(private val db: FirebaseFirestore) : ProfileRep
...
       onSuccess: (Profile) -> Unit,
-      onFailure: (Exception) -> Unit
+      onFailure: (Exception) -> Unit,
+      context: Context
   ) {
     if (fireUser == null) {
       onFailure(Exception("User is null"))
       return
     }
-    val context = ApplicationProvider.getApplicationContext<Context>()
```
Wylwi, line 48: Question: Why is this line removed? What was the issue with it?

Wylwi (commented):
Just a question about a line! 

I also have a more general question, why do you say the addProfile is useless? Can't it be useful later on?

app/src/main/java/ch/hikemate/app/model/profile/ProfileRepositoryFirestore.kt
Wylwi: Oh okay thanks!

rBergCode:
Also, `addProfile` was useless as we now have a `createProfile` function

Wylwi (approved):
Okay, thanks for answering my questions! Looks like a clean refactor! Try to have a better PR and branch name next time 😉

rBergCode:
We will ignore sonarcloud report for this PR as we removes tests but the coverage is still good

PR #159, opened 2024-11-14
feat: improve error handling in profile view model

Wylwi:
# Summary
- Error were not correctly handled in the profile view model, this PR implements an error handling with feedback to the user.

# Details
- I removed default values since they should not have one now, either we have the correct profile, or we don't and it's an error.
- I changed the `errorMessage` to `errorMessageId`, since it's a string ID
- I changed the way the UI handles the update of the profile, in such a way that we can now detect an error before moving the user back
- I removed a navigation test because it just creates an account, without going through the create profile process (Since he does it using direct firebase function), and thus I don't have a profile to load, which throws an error screen. Such test is in reality a E2E test and is already covered by E2E tests

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
@@ -48,7 +48,26 @@ class EndToEndTest2 {
     FirebaseApp.initializeApp(context)
 
     auth.createUserWithEmailAndPassword(email, password)
+
+    var signedOut = false
+
+    // Wait for sign out to complete
+    FirebaseAuth.getInstance().addAuthStateListener {
+      if (it.currentUser == null) {
+        signedOut = true
+      }
+    }
+
     auth.signOut()
+
+    val timeout = System.currentTimeMillis() + 10000 // 10 seconds
+    while (!signedOut && System.currentTimeMillis() < timeout) {
+      Thread.sleep(100)
+    }
+
+    if (!signedOut) {
+      throw Exception("Failed to sign out")
+    }
   }
 
   @After
--- a/app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
@@ -1,22 +1,15 @@
 package ch.hikemate.app.navigation
 
-import android.content.Context
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
-import androidx.compose.ui.test.performTextInput
-import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.hikemate.app.HikeMateApp
 import ch.hikemate.app.ui.auth.SignInScreen
 import ch.hikemate.app.ui.auth.SignInWithEmailScreen
 import ch.hikemate.app.ui.components.BackButton
-import ch.hikemate.app.ui.navigation.Route
 import ch.hikemate.app.ui.navigation.Screen
-import ch.hikemate.app.ui.navigation.TEST_TAG_MENU_ITEM_PREFIX
-import ch.hikemate.app.ui.saved.SavedHikesScreen
-import com.google.firebase.FirebaseApp
 import com.google.firebase.auth.EmailAuthProvider
 import com.google.firebase.auth.FirebaseAuth
 import java.util.UUID
@@ -39,14 +32,6 @@ class HikeMateAppNavigationTest {
   @Before
   fun setUp() {
     composeTestRule.setContent { HikeMateApp() }
-    auth.signOut()
-  }
-
-  private fun setupUser() {
-    val context = ApplicationProvider.getApplicationContext<Context>()
-    FirebaseApp.initializeApp(context)
-
-    auth.createUserWithEmailAndPassword(email, password)
     auth.signOut()
   }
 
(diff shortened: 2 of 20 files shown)


app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -62,16 +67,51 @@ fun EditProfileScreen(
...
   var name by remember { mutableStateOf(profile.name) }
   var hikingLevel by remember {
     mutableIntStateOf(HikingLevel.values().indexOf(profile.hikingLevel))
   }
 
+  var savedProfile by remember { mutableStateOf<Profile?>(null) }
+
+  var isLoading by remember { mutableStateOf(false) }
+
+  if (savedProfile == profile) {
```
rBergCode, line 103: Praise: nice check!

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -157,15 +197,15 @@ fun EditProfileScreen(
...
                   Profile(
                       profile.id,
                       name,
                       profile.email,
                       HikingLevel.values()[hikingLevel],
                       profile.joinedDate)
-              profileViewModel.updateProfile(profile = savedProfile)
-              navigationActions.goBack()
+              profileViewModel.updateProfile(profile = savedProfile!!)
+              isLoading = true
```
rBergCode, line 208: Warning: you set it to true here but never to false. What would happen if there is an error while updating the profile? Wouldn't the loading keep going ?

app/src/main/res/values/strings.xml
```diff
@@ -134,4 +134,9 @@
 
     <!-- Elevation graph -->
     <string name="elevation_graph_loading_label">Loading graph…</string>
+    <string name="an_error_occurred_while_fetching_the_profile">An error occurred while fetching the profile.</string>
+    <string name="an_error_occurred_while_updating_the_profile">An error occurred while updating the profile.</string>
+    <string name="an_error_occurred_while_deleting_the_profile">An error occurred while deleting the profile</string>
```
rBergCode, line 139: Nitpick: could you move them into their correct section?

app/src/main/java/ch/hikemate/app/MainActivity.kt
```diff
@@ -74,10 +74,10 @@ class MainActivity : ComponentActivity() {
 fun HikeMateApp() {
   val navController = rememberNavController()
   val navigationActions = NavigationActions(navController)
-  val firestore = FirebaseFirestore.getInstance()
-  val profileRepository = ProfileRepositoryFirestore(firestore)
-  val profileViewModel = ProfileViewModel(profileRepository)
-  val authViewModel = AuthViewModel(FirebaseAuthRepository(), profileRepository)
+  val profileViewModel: ProfileViewModel = viewModel(factory = ProfileViewModel.Factory)
```
rBergCode, line 77: Praise: nice cleanup

rBergCode (changes requested):
Nice job! Thanks for the work.
I have small remarks and then you can merge.

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
Wylwi: Nop since we are supposed to change to another screen after it

app/src/main/res/values/strings.xml
Wylwi: Done!

Wylwi:
@rBergCode The PR wasn't ready for review. I addressed your concerns anyway, but now you can review it correctly 😉

app/src/androidTest/java/ch/hikemate/app/ui/profile/ProfileScreenTest.kt
```diff
@@ -100,20 +127,29 @@ class ProfileScreenTest : TestCase() {
                   context.getString(R.string.profile_screen_hiking_level_info_expert)
             }))
 
-    val current = LocalDate.now()
-    val currentDateDayOfMonth = current.dayOfMonth
-    val currentDateMonthText = current.month.toString().lowercase()
+    val current = Calendar.getInstance().time
+    val currentDateDayOfMonthText = SimpleDateFormat("d").format(current).toString()
+    val currentDateMonthText = SimpleDateFormat("MMMM").format(current).lowercase()
```
rBergCode, line 132: Question: Why have you changed that?

rBergCode (changes requested):
Nice job! Pretty big PR but very clean and useful! I have some questions/suggestions before you can merge it.

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -42,14 +35,6 @@ class HikeMateAppNavigationTest {
...
 
-  private fun setupUser() {
-    val context = ApplicationProvider.getApplicationContext<Context>()
-    FirebaseApp.initializeApp(context)
-
-    auth.createUserWithEmailAndPassword(email, password)
-    auth.signOut()
-  }
-
   @After
```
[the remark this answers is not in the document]
Wylwi: Yep, you're right!

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
```diff
@@ -48,7 +48,26 @@ class EndToEndTest2 {
...
     auth.signOut()
+
+    val timeout = System.currentTimeMillis() + 10000 // 10 seconds
+    while (!signedOut && System.currentTimeMillis() < timeout) {
+      Thread.sleep(100)
+    }
+
+    if (!signedOut) {
+      throw Exception("Failed to sign out")
+    }
```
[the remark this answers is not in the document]
Wylwi: It is supposed yes!

app/src/androidTest/java/ch/hikemate/app/ui/profile/ProfileScreenTest.kt
Wylwi: Great question! Because of Locale. On my phone, it was written "Novembre" but the test was checking for "November", thus failing. SimpleDateFormat make use of the locale, thus, it fixes the test

rBergCode (approved):
Thanks for the answers and the changes!

PR #163, opened 2024-11-15
Fix/signout is too fast

davidholzwarth:
# Summary

Add a onSuccess callback so we get better control what happens after signout, this fixes the the end2end test failing. Also updated the auth vM tests accordingly.


rBergCode (approved):
LGTM! Thanks for the quick fix

PR #164, opened 2024-11-15
test(end2end): implement second end2end

MaelImhof:
# Summary

Implements a second meaningful end-to-end test that saves a hike in the details screen and checks it has indeed been saved.


davidholzwarth (approved):
LGTM

PR #167, opened 2024-11-16
refactor(details): hike difficulty as enum

MaelImhof:
# Summary

This is a refactoring PR that turns the detailed hike's difficulty property type into an `enum` instead of a `String`. It also makes the name of the difficulty level translatable by moving it to the `strings.xml` file.

Not much more to say, this is a quite straight-forward PR.


rBergCode (approved):
LGTM! Thanks for the quick implementation!

PR #169, opened 2024-11-16
feat(routes): deterministic color for hikes

rBergCode:
# Summary
Implemented a deterministic color for each hikes so that the color won't change.

# Details
For this, I get the hash of the id modulo it to get the index of the color to use. The colors are hardcoded and selected by me so that they look good and enough so that there are not too much crossing of hikes of the same color.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
@@ -138,7 +138,6 @@ data class HikeRoute(
           0xFF7E57C2.toInt(), // Purple
           0xFF1E88E5.toInt(), // Blue
           0xFFFFA726.toInt(), // Orange
-          0xFF000000.toInt(), // Black
           0xFFEF5350.toInt(), // Red
           0xFF26A69A.toInt() // Teal
           )
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
@@ -1,5 +1,6 @@
 package ch.hikemate.app.model.route
 
+import kotlin.math.abs
 import kotlin.math.asin
 import kotlin.math.cos
 import kotlin.math.sin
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -130,6 +131,21 @@ data class HikeRoute(
...
+) {
+
+  private val colors =
+      listOf(
+          0xFF7E57C2.toInt(), // Purple
+          0xFF1E88E5.toInt(), // Blue
+          0xFFFFA726.toInt(), // Orange
+          0xFFEF5350.toInt(), // Red
+          0xFF26A69A.toInt() // Teal
+          )
```
MaelImhof, line 143: Suggestion: could those colors be moved to the `Color.kt` file of the project? Or maybe we have a task for this, I'm not sure.

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -130,6 +131,21 @@ data class HikeRoute(
...
+
+  /** Get the color of the route from its id. The color should be the same for the same route id. */
+  fun getColor(): Int {
+    return colors[abs(id.hashCode()) % colors.size]
+  }
```
MaelImhof, line 148: Praise: very nice way of implementing it, and I like the colors you chose as well !

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -252,8 +254,7 @@ fun HikeDetails(
                       .testTag(TEST_TAG_ELEVATION_GRAPH),
               styleProperties =
                   ElevationGraphStyleProperties(
-                      strokeColor = MaterialTheme.colorScheme.primary,
-                      fillColor = (MaterialTheme.colorScheme.primary).copy(0.1f)))
+                      strokeColor = hikeColor, fillColor = hikeColor.copy(0.1f)))
```
MaelImhof, line 257: Praise: looks even better now with all the colors !

MaelImhof (commented):
Very elegant and nice-looking implementation, the app feels way more "done" now that the colors are actually controlled and systematically the same for a single hike.

Just one comment about the colors, maybe we should put them in `Color.kt`, but maybe that's not even necessary for now, I believe we have a task for the material theme.

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
rBergCode: I will do it now

MaelImhof (approved):
Thanks for the changes, looks perfect to me now !

PR #172, opened 2024-11-16
Fix/rename altitude difference to elevation gain

davidholzwarth:
# Summary

Changed `Altitude difference` in the Map screen to `Elevation gain`, and updated the calculations accordingly.

# Details

A calculation for elevation gain is already made in the DetailedHikeRoute class, so to remove redundancy, I moved the DetailedHikeRoute functions to a separate opject, `RouteUtils`, which now contains five functions which calculate useful information about hikes (elevation gain, estimated time, etc.). Currently, the DetailedHikeRoute uses these Utils for getting the "details" about its hike, and the MapScreen. 

The tests were changed accordingly.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import ch.hikemate.app.R
 import ch.hikemate.app.ui.map.MapScreen
+import ch.hikemate.app.utils.RouteUtils
 
 object HikeCard {
   const val TEST_TAG_HIKE_CARD_TITLE = "HikeCardTitle"
@@ -71,8 +72,7 @@ fun HikeCard(
     showGraph: Boolean = true,
 ) {
   val displayMessage = false
-  val altitudeDifference =
-      if (elevationData.isNullOrEmpty()) null else elevationData.max() - elevationData.min()
+  val elevationGain = RouteUtils.calculateElevationGain(elevationData ?: emptyList())
 
   Row(
       modifier =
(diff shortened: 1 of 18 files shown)


app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -115,12 +115,11 @@ fun HikeCard(
                           style = MaterialTheme.typography.bodySmall)
                       Text(
                           text =
-                              if (altitudeDifference == null)
+                              if (elevationGain == 0.0)
                                   stringResource(R.string.hike_card_no_data_label)
```
rBergCode, line 119: Warning: If a hike has a real `elevationGain` of 0m we don't want to show "no_data". I think we should keep the convention of `elevationGain` to be null if we don't have any data.

app/src/main/res/values/strings.xml
```diff
@@ -6,8 +6,8 @@
     <string name="default_web_client_id">154591189513-eilbr2jtms0av2ses4tsvo62009q5t7n.apps.googleusercontent.com</string>
 
     <!-- Hike Card -->
-    <string name="hike_card_altitude_label">Altitude difference</string>
-    <string name="hike_card_altitude_value_template">%sm</string>
+    <string name="hike_card_altitude_label">Elevation gain</string>
+    <string name="hike_card_altitude_value_template">%.0fm</string>
```
rBergCode, line 10: Nitpick: You should change the name of the strings to reflect the change.

app/src/main/java/ch/hikemate/app/utils/RouteUtils.kt
rBergCode, line 1: Praise: Nice! This is way cleaner and easier to use.

app/src/test/java/ch/hikemate/app/utils/RouteUtilsTest.kt
```diff
@@ -0,0 +1,161 @@
...
+
+    assertEquals(207.59, RouteUtils.computeTotalDistance(points), 0.1)
+  }
+
+  @Test
+  fun test_computeTotalDistance_returns_zero_for_empty_list() {
+    val points = emptyList<LatLong>()
+    val totalDistance = RouteUtils.computeTotalDistance(points)
+    assertEquals(0.0, totalDistance, 0.1)
+  }
```
rBergCode, line 85: Praise: Nice edge case test!

app/src/test/java/ch/hikemate/app/utils/RouteUtilsTest.kt
rBergCode, line 1: Praise: Really nice unit test. Good job!

rBergCode (changes requested):
Good job! Really clean PR. 
I have some suggestions for you to address before merging.

PR #177, opened 2024-11-18
feat(details): limit map scroll and zoom

MaelImhof:
# Summary

The purpose of the details screen is to see more details about a particular hike. To this end, I was assigned to limit the scrolling and zooming the user can do on that screen, so that the map stays around the detailed hike.

# Details

I had to use an event provided by OSMDroid.

When creating the map (`val mapView = remember { ... }`), the center and the initial zoom level are provided to the map. However, the map does not compute its bounding box right away. I listen to the event that gets triggered when the map is ready to be displayed. By this point, the initial bounding box has been computed, and I can limit the scroll to that.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
@@ -78,8 +78,8 @@ import org.osmdroid.views.MapView
 
 object HikeDetailScreen {
   const val MAP_MAX_ZOOM = 18.0
-  const val MAP_MAX_LONGITUDE = 175.0
-  const val MAP_MIN_LONGITUDE = -175.0
+  const val MAP_MAX_LONGITUDE = 180.0
+  const val MAP_MIN_LONGITUDE = -180.0
   const val MAP_BOUNDS_MARGIN: Int = 100
 
   const val TEST_TAG_MAP = "map"


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -151,8 +155,22 @@ fun HikeDetailScreen(
...
+    mapView.setScrollableAreaLimitLatitude(
+        min(MapScreen.MAP_MAX_LATITUDE, mapView.boundingBox.latNorth),
+        max(MapScreen.MAP_MIN_LATITUDE, mapView.boundingBox.latSouth),
+        HikeDetailScreen.MAP_BOUNDS_MARGIN)
+    if (route.bounds.maxLon < HikeDetailScreen.MAP_MAX_LONGITUDE ||
+        route.bounds.minLon > HikeDetailScreen.MAP_MIN_LONGITUDE) {
+      mapView.setScrollableAreaLimitLongitude(
+          max(HikeDetailScreen.MAP_MIN_LONGITUDE, mapView.boundingBox.lonWest),
+          min(HikeDetailScreen.MAP_MAX_LONGITUDE, mapView.boundingBox.lonEast),
+          HikeDetailScreen.MAP_BOUNDS_MARGIN)
```
rBergCode, line 173: Question: How does this manage the case when a hike is on the ±180° line ?

rBergCode (commented):
Nice and quick fix! I have 2 questions which I would need to know about before we can merge it. Thanks!

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -70,14 +69,18 @@ import ch.hikemate.app.utils.toFormattedString
...
 import kotlin.math.roundToInt
 import org.osmdroid.config.Configuration
 import org.osmdroid.views.CustomZoomButtonsController
 import org.osmdroid.views.MapView
 
 object HikeDetailScreen {
-  const val MAP_MIN_ZOOM = 3.0
   const val MAP_MAX_ZOOM = 18.0
+  const val MAP_MAX_LONGITUDE = 175.0
+  const val MAP_MIN_LONGITUDE = -175.0
```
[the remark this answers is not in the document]
MaelImhof: Thought you used those values in one of your previous PRs, but after checking you did it for latitude, not longitude. Changed it.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
MaelImhof: It doesn't. Note however that since the map is meant to be centered on a hike, and the `hike.bounds` attribute is a `Bounds` instance of our own making, it is not possible for a hike to cross that line, by design of the `Bounds` data class.

rBergCode (approved):
LGTM!

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
rBergCode: Okay thanks!

PR #181, opened 2024-11-18
fix: update dateTimeUtils tests to work for non-english locales too

davidholzwarth:
# Summary

- What is the objective? What is fixed/what feature is added/what was documented?
- How is the objective implemented? What classes/doc files/CI were changed?

# Details

Depends on the nature of the PR :
- **Code changes :** describe the conceptual approach taken in the code and tell the reviewer what part of the changes does what, so that it's easier to read the code
- **Other changes :** Add details if relevant, or remove this section if not useful

### DIFF
--- a/app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
+++ b/app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
@@ -32,8 +32,14 @@ fun Timestamp.Companion.from(year: Int, month: Int, day: Int): Timestamp {
   return Timestamp(instant.epochSecond, instant.nano)
 }
 
-fun Timestamp.toLocalDate(): LocalDate {
-  return this.toDate().toInstant().atZone(ZoneId.systemDefault()).toLocalDate()
+fun Timestamp.toLocalDate(locale: Locale? = null): LocalDate {
+
+  // attempts to find the zone for a given locale, useful for injecting locales for test
+  val zone =
+      locale?.country?.takeIf { it.isNotEmpty() }?.let { ZoneId.of("GMT") }
+          ?: ZoneId.systemDefault()
+
+  return this.toDate().toInstant().atZone(zone).toLocalDate()
 }
 
 fun Timestamp.humanReadableFormat(locale: Locale = Locale.getDefault()): String {
@@ -47,7 +53,7 @@ fun Timestamp.humanReadablePlannedLabel(
     locale: Locale = Locale.getDefault(),
     currentDate: LocalDate = LocalDate.now()
 ): String {
-  val plannedDate = this.toLocalDate()
+  val plannedDate = this.toLocalDate(locale = locale)
   val daysDifference =
       Duration.between(currentDate.atStartOfDay(), plannedDate.atStartOfDay()).toDays()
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
```diff
@@ -32,8 +32,14 @@ fun Timestamp.Companion.from(year: Int, month: Int, day: Int): Timestamp {
...
-fun Timestamp.toLocalDate(): LocalDate {
-  return this.toDate().toInstant().atZone(ZoneId.systemDefault()).toLocalDate()
+fun Timestamp.toLocalDate(locale: Locale? = null): LocalDate {
+
+  // attempts to find the zone for a given locale, useful for injecting locales for test
+  val zone =
+      locale?.country?.takeIf { it.isNotEmpty() }?.let { ZoneId.of("GMT") }
+          ?: ZoneId.systemDefault()
+
+  return this.toDate().toInstant().atZone(zone).toLocalDate()
```
MaelImhof, line 42: Changes: Two things here:

- Passing a parameter is nice, however a locale cannot necessarily be mapped to a time zone. Could we simply pass a `ZoneId` instead? It will only be used for tests anyway.
- I realize I made a mistake when implementing the original function, I used `this.toDate().toInstant()`, but `Timestamp.toInstant()` already exists, so we can use `this.toInstant()` instead.

```suggestion
fun Timestamp.toLocalDate(zoneId: ZoneId?): LocalDate {
  return this.toInstant().atZone(zoneId).toLocalDate()
}
```

app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
```diff
@@ -47,7 +53,7 @@ fun Timestamp.humanReadablePlannedLabel(
     locale: Locale = Locale.getDefault(),
     currentDate: LocalDate = LocalDate.now()
 ): String {
-  val plannedDate = this.toLocalDate()
+  val plannedDate = this.toLocalDate(locale = locale)
```
MaelImhof, line 56: Suggestion: we should probably not pass anything here, so that the default system value is used. Mapping a locale to a zone id is hard to get right, we can leave it to the system.

```suggestion
  val plannedDate = this.toLocalDate()
```

app/src/test/java/ch/hikemate/app/utils/DatetimeUtilsTest.kt
MaelImhof, line 1: Praise: thank you for pointing this out, did not realize my tests depended on the locale of my system.

MaelImhof (changes requested):
Very nice catch ! A few changes that I think we should fix differently, but other than that very nice fix.

app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
davidholzwarth: Thanks for the suggestion. While going over your review I noticed that, errenously, I had assumed that this function also changes the format of the date, however that is not the case, so the following suffices:
```kotlin 
fun Timestamp.toLocalDate(zoneId: ZoneId = ZoneId.systemDefault()): LocalDate {
  return this.toInstant().atZone(zoneId).toLocalDate()
}

MaelImhof (approved):
Very nice, seems good to me. I checked and the tests still pass locally for me, they seem to pass on the CI too. As long as they work for you locally, this PR looks ready to me. Nice work !

PR #183, opened 2024-11-18
fix:aesthetic fixes for hike details screen

davidholzwarth:
# Summary

- Changed `00:00` to `00h00` on the hike details screen, and made it be 00min if the time is less than an hour
- Added changing color to hike difficulty level on hike details screen
- Added a more dark-more friendly implementation of the bookmark icon

# Details

For the color, I expanded the `HikeDifficulty` enum to also contain a color Id, and added corresponding colors to the resources. The colors (green, yellow and orange) were chosen such that visibility is fine in both light and dark mode.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
@@ -54,7 +54,6 @@ import ch.hikemate.app.ui.components.BackButton
 import ch.hikemate.app.ui.components.ElevationGraph
 import ch.hikemate.app.ui.components.ElevationGraphStyleProperties
 import ch.hikemate.app.ui.map.HikeDetailScreen.MAP_MAX_ZOOM
-import ch.hikemate.app.ui.map.HikeDetailScreen.MAP_MIN_ZOOM
 import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_ADD_DATE_BUTTON
 import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_BOOKMARK_ICON
 import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_DETAIL_ROW_TAG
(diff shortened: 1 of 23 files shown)


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -292,11 +293,15 @@ fun HikeDetails(
               value = "${elevationGainString}m")
           DetailRow(
               label = stringResource(R.string.hike_detail_screen_label_estimated_time),
-              value = "${hourString}:${minuteString}")
+              value = "${hourString}h${minuteString}")
```
mchac15, line 296: Question: I find it weird that when the hike is less than one hour 00h25m is displayed. Wouldn't it be better to display 25m instead? Do you know if it is possible to do so?

app/src/main/res/values/colors.xml
```diff
@@ -3,4 +3,8 @@
     <color name="blue">#FF3B9DE8</color>
     <color name="black">#FF000000</color>
     <color name="white">#FFFFFFFF</color>
+
+    <color name="hike_difficulty_easy">#1B9E4D</color>
+    <color name="hike_difficulty_moderate">#FCCA03</color>
+    <color name="hike_difficulty_difficult">#F74E05</color>
```
mchac15, line 9: Praise: Good practice.

mchac15 (changes requested):
I found one little thing you could change still. While in Dark mode when the hike is not saved the _Not Saved_ info is black and in contrast with the drawer is almost impossible to change.
Also don't forget to change the title of the PR to fit the convention.
fix: aesthetic fixes for hike details screen

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
davidholzwarth: Good point, thats not too difficult to implement. I've changed it to be 00min if the time is less than an hour

davidholzwarth:
@mchac15 About the dark mode, thats a good point too, if you check now the Icon has recieved a grey border instead of a black one which is easily visible in dark and light mode, and the text changes color depending on dark/light mode, its no longer hardcoded black.

mchac15 (approved):
Nice work!
I like the subtle changes and thanks for listening to the feedback.

PR #184, opened 2024-11-18
ci: add jacoco code coverage to the ci

Wylwi:
Sonar is working on top of the Jacoco report, but doesn't compute it the same way. This upload the report to github actions. 

This PR also adds a small jacoco parser, to show more detailed description in the PR, without having to go all the way to the zip file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,30 @@ jobs:
       # Upload the various reports to sonar
       - name: Upload report to SonarCloud
         env:
-          GITHUB_TOKEN: ${{ secrets.TOKEN_GITHUB }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
         run: ./gradlew sonar --parallel --build-cache
 
+      - name: Jacoco Report to PR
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -127,7 +127,30 @@ jobs:
       # Upload the various reports to sonar
       - name: Upload report to SonarCloud
         env:
-          GITHUB_TOKEN: ${{ secrets.TOKEN_GITHUB }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rBergCode, line 130: Question: why was this changed ?

.github/workflows/ci.yml
rBergCode, line 1: Praise: nice ! The message looks good!

rBergCode (commented):
I have one question and you'll be good to go! Nice job!

.github/workflows/ci.yml
Wylwi: I initially created a TOKEN_GITHUB secret with a handmade token, but GitHub, when launching the CI, gives a GITHUB_TOKEN by default, which is a secret specifically created for this specific CI run. This is just a security improvement

rBergCode (approved):
Thanks!

PR #188, opened 2024-11-20
ci: make jacoco code coverage requirement only enforce on main

Wylwi:
Jacoco report fails when not running on a PR, since he can't add a comment. This makes it run the jacoco report only on PR (Code coverage upload still works). This changes also make it such that the CI runs on every branch, even when there is no PR open

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - main
-      - "*"
+      - "**"
 
   pull_request:
     types:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -142,14 +149,9 @@ jobs:
...
 
-      - name: Upload Code Coverage Artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: Code-coverage-report
-          path: ${{ github.workspace }}/app/build/reports/jacoco/jacocoTestReport/
 
       - name: Fail PR if overall coverage is less than 80%
-        if: ${{ steps.jacoco.outputs.coverage-overall < 80.0 }}
+        if: ${{ startsWith(github.event_name, 'pull_request') && steps.jacoco.outputs.coverage-overall < 80.0 }}
```
MaelImhof, line 154: Question: does this condition mean that the CI will succeed if just on a branch (not on a PR) and there is less than 80% code coverage? Is there any way to fail on branches too if there's missing coverage?

MaelImhof (commented):
Great fix overall! Just have one question about how the CI will run (and fail) on branches.

.github/workflows/ci.yml
MaelImhof: Fair, having a fail in the PR is sufficient.

Issue #124, opened 2024-11-08
Implement delete account


rBergCode:
Currently, deleting the user from firebase doesn't delete his data in the database. This will be an issue for tests and real app usage. Doing it will also improve maintainability.

One need to alter every test deleting the user with the above created function and any Model that store data.

Acceptance criteria : 
- [ ] All tests that create and delete accounts should use this method
- [ ] When deleting an account, the database should be emptied from user's data (saved hikes, profile data)

rBergCode:
This was more difficult than expected. as this is not very important we will do it in a later spring

PR #189, opened 2024-11-20
feat(auth): deleting a user from the ui

rBergCode:
# Summary
The objective was to implement the delete account feature to let user delete their account.

# Details
I also did some useful changes:

- Move the collection name of the profile repository to a companion object.
- Implemented a `isEmailProvider` to know wether the user is connected via email or google
- Kind of fixed reliability of the `testWholeNavigation` test


app/src/main/java/ch/hikemate/app/model/authentication/AuthRepository.kt
```diff
@@ -71,4 +72,29 @@ interface AuthRepository {
...
+      onSuccess: () -> Unit = {},
+      onErrorAction: (Exception) -> Unit
+  )
+
+  /**
+   * Return true if the user is signed in with an email provider.
+   *
+   * @param user The FirebaseUser object.
+   */
+  fun isEmailProvider(user: FirebaseUser): Boolean
```
Wylwi, line 99: Nitpick: Sorry about this, really nitpicking. Could you elaborate about what is an "email provider"? Strictly speaking, Facebook Auth provides email, so does Google, and so on... It might be clear when you're doing this, but for a newcomer, it's not easily understandable

app/src/main/java/ch/hikemate/app/model/authentication/AuthViewModel.kt
```diff
@@ -148,4 +149,36 @@ class AuthViewModel(
...
+        activity = activity,
+        onSuccess = {
+          _currentUser.value = null
+          onSuccess()
+        },
+        onErrorAction = onErrorAction)
+  }
+
+  /** Returns whether the current user is connect with an email provider. */
+  fun isEmailProvider(): Boolean {
```
Wylwi, line 177: Ditto: Same as above about email provider

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -127,4 +134,91 @@ class FirebaseAuthRepository : AuthRepository {
...
+
+            // Delete the user profile and saved hikes
+            FirebaseFirestore.getInstance()
+                .collection(ProfileRepositoryFirestore.PROFILES_COLLECTION)
+                .document(user.uid)
+                .delete()
+            FirebaseFirestore.getInstance()
+                .collection(SavedHikesRepositoryFirestore.SAVED_HIKES_COLLECTION)
+                .document(user.uid)
+                .delete()
```
Wylwi, line 165: Question: Do you have an easy way to check for correctness of the data deletion? I'm pretty sure Firebase should return a boolean somewhere to indicate that deletion has correctly happened

Wylwi: Part of my concern is about the deletion of the Firebase user (and thus permission) before the deletion is finished, leading to permissions errors

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -127,4 +134,91 @@ class FirebaseAuthRepository : AuthRepository {
...
+          },
+          { onErrorAction(Exception("Error re-authenticating user")) })
+    } else {
+      onErrorAction(Exception("No user is currently signed in"))
+    }
+  }
+
+  override fun isEmailProvider(user: FirebaseUser): Boolean {
+    val providers = user.providerData.map { it.providerId }
+    return providers.contains(EmailAuthProvider.PROVIDER_ID)
```
Wylwi, line 187: Nitpicking: Don't you have a more performance efficient way to do it? Currently, this implementation is θ(n^2) in CPU usage (not accounting for comparison). I'm confident you can check for this in a θ(n) way

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -127,4 +134,91 @@ class FirebaseAuthRepository : AuthRepository {
...
+          .addOnFailureListener { e ->
+            Log.e("DeleteAccount", "Re-authentication failed: ${e.message}")
+            onFailure()
+          }
+          .addOnSuccessListener { result ->
+            Log.d("DeleteAccount", "Re-authentication successful: $result")
+            onSuccess()
+          }
+    }
+  }
```
Wylwi, line 223: Praise: Nice and efficient way to do a reauthentication!

app/src/main/java/ch/hikemate/app/model/profile/ProfileRepositoryFirestore.kt
Wylwi, line 1: Praise: Nice refactor!

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
```diff
@@ -0,0 +1,94 @@
...
+  private lateinit var navigationActions: NavigationActions
+
+  @get:Rule val composeTestRule = createComposeRule()
+
+  @Before
+  fun setUp() {
+    navigationActions = mockk(relaxed = true)
+    authViewModel = mockk(relaxed = true)
+
+    every { navigationActions.goBack() } returns Unit
```
Wylwi, line 30: Question: Why is this needed? Is it an issue if we go back?

app/src/test/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
```diff
@@ -304,4 +336,48 @@ class FirebaseAuthRepositoryTest {
...
+      every { mockVoidTask.isSuccessful } returns true
+      every { mockFirebaseAuth.currentUser } returns mockFirebaseUser
+      every { mockFirebaseUser.providerData } returns
+          listOf(mockk { every { providerId } returns EmailAuthProvider.PROVIDER_ID })
+      every { mockDocument.delete() } returns mockVoidTask
+
+      val onSuccess: () -> Unit = mockk(relaxed = true)
+      val mockActivity: Activity = mockk(relaxed = true)
+
+      repository.deleteAccount("password", mockActivity, onSuccess, mockOnError)
```
Wylwi, line 352: Question: Do we agree that in this setup (i.e. with those mocks), "password" will never be checked for correctness?

app/src/test/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
Wylwi, line 1: Praise: You did an amazing job on testing this repo!!

Wylwi, line 1: Nitpick: Could you add one or two tests where the tests check for failure please?

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
```diff
@@ -361,4 +362,50 @@ class AuthViewModelTest {
...
+    doAnswer { arguments ->
+          val onSuccess = arguments.getArgument<() -> Unit>(2)
+          onSuccess()
+          null
+        }
+        .`when`(mockRepository)
+        .deleteAccount(any(), any(), any(), any())
+
+    // Verify that currentUser is initially mockFirebaseUser
+    assertEquals(mockFirebaseUser, viewModel.currentUser.first())
```
Wylwi, line 382: Question: Why do you use `.first()` here and not `value`? `.first()`, even if it might be working, doesn't look appropriated for a single user

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
```diff
@@ -361,4 +362,50 @@ class AuthViewModelTest {
...
+    setupSignedInUser()
+
+    // Simulate a successful call to isEmailProvider
+    `when`(mockRepository.isEmailProvider(any())).thenReturn(true)
+
+    viewModel.isEmailProvider()
+
+    // Verify that the repository's isEmailProvider was called
+    verify(mockRepository).isEmailProvider(any())
+  }
```
Wylwi, line 410: Praise: Simple but efficient test!

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
Wylwi, line 1: Nitpick: Could you add one or two tests where the tests check for failure please?

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
```diff
@@ -0,0 +1,94 @@
...
+  @Test
+  fun isEverythingDisplayed() {
+    setUpActivityEmailProvider()
+    composeTestRule.onNodeWithTag(BackButton.BACK_BUTTON_TEST_TAG).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(DeleteAccountScreen.TEST_TAG_TITLE).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(DeleteAccountScreen.TEST_TAG_INFO_TEXT).assertIsDisplayed()
+    composeTestRule
+        .onNodeWithTag(DeleteAccountScreen.TEST_TAG_DELETE_ACCOUNT_BUTTON)
+        .assertIsDisplayed()
+  }
```
Wylwi, line 52: Warning: This doesn't check for `TEST_TAG_PASSWORD_INPUT`

app/src/main/java/ch/hikemate/app/ui/profile/DeleteAccountScreen.kt
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @param navigationActions The navigation actions.
+ * @param authViewModel The authentication view model.
+ */
+@Composable
+fun DeleteAccountScreen(navigationActions: NavigationActions, authViewModel: AuthViewModel) {
+  val context = LocalContext.current
+
+  val passwordMustNotBeEmptyError =
+      stringResource(R.string.delete_account_password_must_be_filled_error)
```
Wylwi, line 55: Question: Why do you have this string here and all the other ones in the composable tree? Could you maybe normalize it, in the way you want?

app/src/main/java/ch/hikemate/app/ui/profile/DeleteAccountScreen.kt
```diff
@@ -0,0 +1,120 @@
...
+                  TextSelectionColors(
+                      handleColor = primaryColor,
+                      backgroundColor = primaryColor,
+                  ))
+
+  var password by remember { mutableStateOf("") }
+
+  Column(
+      modifier =
+          Modifier.testTag(Screen.SIGN_IN_WITH_EMAIL)
```
Wylwi, line 74: Error: You put the wrong test tag here

app/src/main/java/ch/hikemate/app/ui/profile/DeleteAccountScreen.kt
```diff
@@ -0,0 +1,120 @@
...
+
+        if (authViewModel.isEmailProvider())
+            OutlinedTextField(
+                modifier =
+                    Modifier.fillMaxWidth().testTag(DeleteAccountScreen.TEST_TAG_PASSWORD_INPUT),
+                visualTransformation = PasswordVisualTransformation(),
+                colors = inputColors,
+                value = password,
+                onValueChange = { password = it },
+                label = { Text(stringResource(R.string.delete_account_password_label)) })
```
Wylwi, line 101: Praise: Clean way to do it!

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
Wylwi, line 1: Question: Why is this file in the profile folder and not in the auth folder? Is it arbitrary or is there any reason?

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
```diff
@@ -0,0 +1,94 @@
...
+  @Test
+  fun checkCanDeleteAccount() {
+    setUpActivityEmailProvider()
+
+    composeTestRule
+        .onNodeWithTag(DeleteAccountScreen.TEST_TAG_PASSWORD_INPUT)
+        .performTextInput("password")
+    composeTestRule.onNodeWithTag(DeleteAccountScreen.TEST_TAG_DELETE_ACCOUNT_BUTTON).performClick()
+
+    verify(exactly = 1) { authViewModel.deleteAccount(any(), any(), any(), any()) }
```
Wylwi, line 83: Suggestion: This is not required, but I think it would be a great improvement to check that the authViewModel get called with the correct password.
Below is a suggestion done without testing. Syntax might not be correct and the "password" should be put in a constant
```suggestion
    verify(exactly = 1) { authViewModel.deleteAccount(eq("password"), any(), any(), any()) }
    verify(exactly = 0) { authViewModel.deleteAccount(not(eq("password")), any(), any(), any()) }
```

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Test
+  fun checkCanDeleteAccountWithoutPassword() {
```
Wylwi, line 87: Nitpick: The name of the test is (in my understanding) not corresponding to the actual behavior of the test

Wylwi (changes requested):
I'm not going to lie, you did an awesome job on this. I am just asking for a few changes to be done, some nitpicking and a lot of questions! Your work is impressive overall!

app/src/androidTest/java/ch/hikemate/app/navigation/HikeMateAppNavigationTest.kt
```diff
@@ -38,8 +38,8 @@ class HikeMateAppNavigationTest {
 
   @Before
   fun setUp() {
-    composeTestRule.setContent { HikeMateApp() }
     auth.signOut()
+    composeTestRule.setContent { HikeMateApp() }
```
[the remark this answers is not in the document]
rBergCode: I did this as I though that it would make the tests pass more accurately. Actually it doesn't but in #159 you will remove this test anyway so I will merge it as soon as you merge it to main!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
rBergCode: Actually it's θ(n) not θ(n^2) but I will change it for θ(n)

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
rBergCode: That's true! It's because the password field would not appear if the user is signed in with Google. That's why I added 2 more tests to test this only. Do you think I should still add it here ?

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
rBergCode: It is as the delete process is part of the profile page (from the `ProfileScreen` -> `DeleteAccountScreen`)

app/src/androidTest/java/ch/hikemate/app/ui/profile/DeleteAccountScreenTest.kt
rBergCode: The issue is because as we set the content of the test activity to be our `DeleteAccountScreen`, there are no previous page. Going back would then throw an error

app/src/test/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
rBergCode: What do you mean by "will never be checked for correctness? It won't do any request to Firebase for sure. So If i understand your question correctly I would say yes.

app/src/test/java/ch/hikemate/app/authentication/AuthViewModelTest.kt
rBergCode: Actually, I took this from this file's others tests. I will change them all to `.value`. Nice catch!

app/src/main/java/ch/hikemate/app/ui/profile/DeleteAccountScreen.kt
rBergCode: It is because this string is used inside of a click function which is not in the composable tree. All others are.
Do you think I should put them all in the same place ?

app/src/main/java/ch/hikemate/app/model/authentication/AuthRepository.kt
```diff
@@ -71,4 +72,29 @@ interface AuthRepository {
...
+   * @param password The password of the user.
+   * @param activity The Android Activity, used for launching the re-authentication dialog.
+   * @param onSuccess Callback to invoke after the user's account has been successfully deleted.
+   * @param onErrorAction Callback to invoke when an error occurs during account deletion. Passes
+   *   the Throwable error.
+   */
+  fun deleteAccount(
+      password: String,
+      activity: Activity,
+      onSuccess: () -> Unit = {},
```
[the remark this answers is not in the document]
rBergCode: It was a mistake. Good catch!

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
rBergCode: I fixed this. Note that there is a sonar warning about `Task<Void>` for which it suggest to change it to `Task<Unit>`. This doesn't work as the tasks types for firebase operations are `Task<Void>`. I suggest that we remove this check in sonar as this is not that bad of a practice.

Wylwi:
@rBergCode I advise you resolve the conflict before I review it again

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
```diff
@@ -155,24 +157,33 @@ class FirebaseAuthRepository : AuthRepository {
...
+                    .delete()
+
+            val deleteSavedHikesTask =
+                FirebaseFirestore.getInstance()
+                    .collection(SavedHikesRepositoryFirestore.SAVED_HIKES_COLLECTION)
+                    .document(user.uid)
+                    .delete()
+
+            Tasks.whenAll(deleteProfileTask, deleteSavedHikesTask).addOnCompleteListener {
+                deleteDataTask: Task<Void?> ->
```
Wylwi, line 173: Warning: Even if Firebase operations are `Task<Void>`, it's still bad practice for both reasons that Sonar gives us : 
1. Wrong logic
   - Void is not equivalent to Java void but `java.lang.Void.`
2. Unnecessary platform dependency
   -  Void is a platform type available only in the Java Runtime Environment.

Second one is more than an issue for us since we are not running using the JRE but ART, which is a specifically designed runtime for Android (ART is simply short for Android RunTime). Thus, declaring `Task<Void>` or any kind of Void is an issue on non-JRE devices. The void keyword is, here, not related to Firebase tasks. If you look carefully at `.whennAll` documentation, it says that it will always return a `Task<Void>`, which is then handled by the library for the correctness of the implementation. 

Since you don't create an object, and just reuse the library created and implemented ones, there exist a simple fix. Simply remove the type annotation, which is not needed here! Then leave it in the care of the library!

Wylwi (changes requested):
We're on the right track! One last change is needed to fix the Sonar warning, and then this PR can be merged!

Wylwi:
Closes #124

app/src/test/java/ch/hikemate/app/authentication/FirebaseAuthRepositoryTest.kt
rBergCode: Then it doesn't check.

app/src/main/java/ch/hikemate/app/model/authentication/FirebaseAuthRepository.kt
rBergCode: Thanks for the explanation and the quick fix!

Wylwi (approved):
Thanks for the quick fix! You can merge it when the checks are finished running!

Issue #137, opened 2024-11-12
Refactor the MapScreen


mchac15:
The `MapScreen` has become too complicated and it needs some refractoring.

- [ ] Refractor the MapScreen into separate chunks of code.

MaelImhof:
> Solve the "bug" in which when a user tries to center the map around the user the animation blocks the map from moving. This should be handled like in Google Maps where the user can cancel this animation and the map won't be centered at the end if the user moves the map.

This is not refactoring, this is a bug fix (or feature depending on how you see it), I think it should be another separate task.

Also, I would add refactoring the location logic, because it needs to be moved to a view model or something, be more structured and maintanable.

PR #193, opened 2024-11-21
feat(routes): display the appropriateness message

rBergCode:
# Summary
The goal was to have a message displaying and telling the user if the hike is suitable or not for him.

# Details
As I added this feature, multiple things had to be done:
- Needed for the profile informations in `MapScreen` and `HikeDetailScreen`
- Needed to fix tests using those screens (as they now need a valid profile
- Needed to fix end-to-end 2 test as it was not creating a profile and still going to map screen

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
@@ -12,6 +12,7 @@ import androidx.compose.ui.test.performTextInput
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.hikemate.app.MainActivity
+import ch.hikemate.app.ui.auth.CreateAccountScreen
 import ch.hikemate.app.ui.auth.SignInScreen
 import ch.hikemate.app.ui.auth.SignInWithEmailScreen
 import ch.hikemate.app.ui.components.BackButton
@@ -46,8 +47,6 @@ class EndToEndTest2 {
   fun setUpFirebase() {
     val context = ApplicationProvider.getApplicationContext<Context>()
     FirebaseApp.initializeApp(context)
-
-    auth.createUserWithEmailAndPassword(email, password)
 
     var signedOut = false
 
@@ -94,14 +93,23 @@ class EndToEndTest2 {
 
     // Perform sign in with email and password
     composeTestRule.onNodeWithTag(SignInScreen.TEST_TAG_SIGN_IN_WITH_EMAIL).performClick()
-    composeTestRule.onNodeWithTag(Screen.SIGN_IN_WITH_EMAIL).assertIsDisplayed()
+
     composeTestRule
-        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_EMAIL_INPUT)
-        .performTextInput(email)
+        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_GO_TO_SIGN_UP_BUTTON)
+        .performClick()
+    composeTestRule.onNodeWithTag(Screen.CREATE_ACCOUNT).assertIsDisplayed()
+
     composeTestRule
-        .onNodeWithTag(SignInWithEmailScreen.TEST_TAG_PASSWORD_INPUT)
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_NAME_INPUT)
+        .performTextInput(myUuidAsString)
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT).performTextInput(email)
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_PASSWORD_INPUT)
         .performTextInput(password)
-    composeTestRule.onNodeWithTag(SignInWithEmailScreen.TEST_TAG_SIGN_IN_BUTTON).performClick()
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_CONFIRM_PASSWORD_INPUT)
+        .performTextInput(password)
+    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
 
     // Wait for the sign-in to be performed and the map to load
     composeTestRule.waitUntilExactlyOneExists(
--- a/app/src/main/java/ch/hikemate/app/ui/components/AsyncStateHandler.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/AsyncStateHandler.kt
@@ -7,8 +7,19 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.State
 import androidx.compose.ui.res.stringResource
 
+/**
+ * A component that handles the state of an asynchronous operation and displays the appropriate
+ * content based on the state. If an error occurred, an error message is displayed. If the value is
+ * null, a loading animation is displayed. Otherwise, the content is displayed.
+ *
+ * @param errorMessageIdState The state that holds the error message id
+ * @param actionContentDescriptionStringId The string id for the content description of the action
+ * @param actionOnErrorAction The action to perform when the error action is clicked
+ * @param valueState The state that holds the value to display
+ * @param content The content to display
+ */
 @Composable
-fun <T> GenericErrorHandler(
+fun <T> AsyncStateHandler(
     errorMessageIdState: State<Int?>,
     actionContentDescriptionStringId: Int,
     actionOnErrorAction: () -> Unit,
--- a/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
@@ -1,5 +1,6 @@
 package ch.hikemate.app.ui.components
 
+import android.util.Log
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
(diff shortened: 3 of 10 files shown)


app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
```diff
@@ -47,8 +48,6 @@ class EndToEndTest2 {
     val context = ApplicationProvider.getApplicationContext<Context>()
     FirebaseApp.initializeApp(context)
 
-    auth.createUserWithEmailAndPassword(email, password)
```
mchac15, line 50: Question: Why was this removed?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -180,28 +192,58 @@ fun HikeDetailScreen(
...
-        onZoomOut = { mapView.controller.zoomOut() },
-        modifier =
-            Modifier.align(Alignment.BottomEnd)
-                .padding(bottom = MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT + 8.dp))
-
-    // Hike Details bottom sheet
-    HikeDetails(detailedRoute, savedHikesViewModel, elevationData)
+  LaunchedEffect(Unit) {
+    if (authViewModel.currentUser.value == null) {
+      Log.e("MapScreen", "User is not signed in")
```
mchac15, line 197: Nitpick: Shouldn't the log message say `HikeDetailScreen` here?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -363,64 +393,90 @@ fun MapScreen(
...
-                onClick = {
-                  MapScreen.launchSearch(isSearching, hikingRoutesViewModel, mapView, context)
-                },
-                modifier =
-                    Modifier.align(Alignment.BottomCenter)
-                        .padding(bottom = MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT + 8.dp))
+  val errorMessageIdState = profileViewModel.errorMessageId.collectAsState()
+  val profileState = profileViewModel.profile.collectAsState()
+
+  when {
```
mchac15, line 399: Question: Since you are reusing this pattern twice on the MapScreen aswell as in the HikeDetailsScreen shouldn't this be made into a function? Or is that not possible?

mchac15 (approved):
The code seems very good, I don't see any major problems with it but for the duplication of code.

mchac15 (changes requested):
I don't see any major problems but for the duplication of code.
Other than that your code is very well structured and very well written

mchac15:
> I don't see any major problems but for the duplication of code. Other than that your code is very well structured and very well written

I forgot to mention that the Sonar major issue about the MapScreen being too complex will be aproved since this will be done in the refactoring of the MapScreen #137

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest2.kt
rBergCode: Good question! I changed the test to create the user in the UI instead of creating it like this.
I changed it because we need to properly create a profile for this to work.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: I think that is the issue with Sonar.
I will try to refactor it to some kind of function but this might be hard to do.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
rBergCode: You're right!

rBergCode:
Those tests don't pass because of either End2end or End2End2 test. I don't know why and can't replicate it in local. @Wylwi have you any ideas ?

Wylwi:
Since you're the only one affected by those errors, no sorry, I don't have any idea... It might be your code, but I don't know which part of it

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
mchac15: Did you try to do it?
Is it too difficult/lengthy to implement?

rBergCode: I just did it!

rBergCode:
The tests were fixed by #197. @mchac15 it should be ready now!

app/src/main/java/ch/hikemate/app/ui/components/AsyncStateHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+ * content based on the state. If an error occurred, an error message is displayed. If the value is
+ * null, a loading animation is displayed. Otherwise, the content is displayed.
+ *
+ * @param errorMessageIdState The state that holds the error message id
+ * @param actionContentDescriptionStringId The string id for the content description of the action
+ * @param actionOnErrorAction The action to perform when the error action is clicked
+ * @param valueState The state that holds the value to display
+ * @param content The content to display
+ */
+@Composable
```
mchac15, line 21: Comment: This function seems pretty good, I've tested again on my phone and by reviewing the code it seems equivalent to the code you had before, but much better, since there isn't any code repetition.
You also did well by treating the cases whenver the user is not signed in and whenever there is an error this makes debuging much easier.

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -71,7 +72,10 @@ fun HikeCard(
     styleProperties: HikeCardStyleProperties = HikeCardStyleProperties(),
     showGraph: Boolean = true,
 ) {
-  val displayMessage = false
+  val displayMessage = !elevationData.isNullOrEmpty() && messageContent != null
+  Log.i(
```
mchac15, line 76: Comment: Nice again, of adding a Log (I didn't comment this before.)

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -185,29 +193,47 @@ fun HikeDetailScreen(
...
-        onZoomIn = { mapView.controller.zoomIn() },
-        onZoomOut = { mapView.controller.zoomOut() },
-        modifier =
-            Modifier.align(Alignment.BottomEnd)
-                .padding(bottom = MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT + 8.dp))
-
-    // Hike Details bottom sheet
-    HikeDetails(detailedRoute, savedHikesViewModel, elevationData)
+  LaunchedEffect(Unit) {
+    if (authViewModel.currentUser.value == null) {
```
mchac15, line 197: Question: Is this case actually possible? If such wouldn't it be more appropriate to show a toast?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -185,29 +193,47 @@ fun HikeDetailScreen(
...
+      Log.e("HikeDetailScreen", "User is not signed in")
+      return@LaunchedEffect
+    }
+    profileViewModel.getProfileById(authViewModel.currentUser.value!!.uid)
+  }
+
+  val errorMessageIdState = profileViewModel.errorMessageId.collectAsState()
+  val profileState = profileViewModel.profile.collectAsState()
+
+  AsyncStateHandler(
```
mchac15, line 207: Praise: The function just makes the code much more readable and it reduces the complexity a bit.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -210,16 +215,31 @@ fun clearHikesFromMap(mapView: MapView, userLocationMarker: Marker?) {
...
+/**
+ * Properties to initialize the map with. These properties are used to set the initial state of the
+ * map, such as the zoom level, the center, and the limits of the zoom.
+ *
+ * @param mapInitialZoomLevel The initial zoom level of the map
+ * @param mapMaxZoomLevel The maximum zoom level of the map
+ * @param mapMinZoomLevel The minimum zoom level of the map
+ * @param mapInitialCenter The initial center of the map
+ */
+data class MapInitialValues(
```
mchac15, line 227: Comment: I like this improvement, but is there a reason why you changed it? I mean a reason related to the PR?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -246,17 +266,23 @@ fun MapScreen(
       // Maximum number of bytes that can be used by the tile file system cache. Default is 600MB.
       tileFileSystemCacheMaxBytes = 600L * 1024L * 1024L
     }
+
+    if (authViewModel.currentUser.value == null) {
```
mchac15, line 270: Same question as in the `HikeDetailsScreen`

mchac15 (changes requested):
The code is very well written and seems to be much improved with the provided function. I added some comments that I didn't do last time.
But overall I don't think there is any major problems and I have tested on my phone looking at the logs and everything seems fine.
Thanks for the changes!

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: It's not really related but I had Sonar complains about to many arguments for MapScreen function

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
rBergCode: It should never happen. This is just a sanity check to prevent NullPointerException

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: Answered above (will resolve it)

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
mchac15: Ok yes that seems more logical

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
mchac15: Good improvemet

mchac15 (approved):
Thanks for answering my questions!
You can merge now

PR #194, opened 2024-11-21
feat: implement model and view model for retrieving trashcans restrooms and parkings

davidholzwarth:
# Summary

This PR implements a viewModel and a model to retrieve facilities to display on a map. For this a Facility class was created, which saved the FacilityType (an enum that was created), and the facility's `GeoPoint` coordinates.

# Details - OverpassRepository

This new class contains helpful constants and functions to remove the redundancy that appeared between the facilityRepository and the hikeRoutesRepository, which both make calls to the Overpass API. 

# Details - FacilitiesRepository

The repository makes an API Call to the Overpass API, parses the result and from these parsed results create a list of facilities. It's getFacilities() function takes bounds, an onFailure and an onSuccess parameter, which allows the caller to apply a function to the created list of Facilities.

# Details - FacilitiesViewModel

The viewModel asynchronously calls the repository to getFacilities for given bounds. The viewModel also has a cache feature, which prevents calls being made to the repository if the given bounds are contained within bounds that have been called previously.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -174,6 +174,9 @@ dependencies {
     implementation(libs.googleid)
     implementation(libs.accompanist.permissions)
     implementation(libs.play.services.location)
+    implementation(libs.gson)
+
+
 
 
     // Global test dependencies
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -177,8 +177,6 @@ dependencies {
     implementation(libs.gson)
 
 
-
-
     // Global test dependencies
     globalTestImplementation(libs.androidx.junit)
     globalTestImplementation(libs.androidx.espresso.core)
--- a/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
+++ b/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
@@ -1,0 +1,103 @@
+package ch.hikemate.app.model.facilities
+
+import android.util.JsonReader
+import android.util.Log
+import ch.hikemate.app.model.route.Bounds
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.osmdroid.util.GeoPoint
+
+data class Facility(val type: String, val coordinates: GeoPoint)
+
+class FacilitiesRepository(private val client: OkHttpClient) {
+
+  fun getFacilities(
+      bounds: Bounds,
+      onSuccess: (List<Facility>) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+
+    val listOfAmenities =
+        "toilets|parking|waste_basket|supermarket|drinking_water|ranger_station|bbq|bench|restaurant|biergarten"
+    val requestData =
+        """
+        [out:json];
+        (
+          node["amenity"~"$listOfAmenities"](${bounds.toStringForOverpassAPI()});        
+        );
+        out geom;
+    """
+            .trimIndent()
+
+    val requestBuilder =
+        Request.Builder().url("https://overpass-api.de/api/interpreter?data=$requestData").get()
+
+    val response = client.newCall(requestBuilder.build()).execute()
+
+    try {
+      if (response.isSuccessful) {
+        val responseBody = response.body?.string() ?: throw Exception("Empty response body")
+
+        val facilities = filterAllAmenities(responseBody)
+
+        Log.d("FacilitiesRepository", "Got facilities: $facilities")
+
+        onSuccess(facilities)
+      }
+    } catch (e: Exception) {
+      Log.d("FacilitiesRepository", "Failed to fetch facilities: ${e.message}")
+      onFailure(e)
+    }
+  }
+
+  // Function to filter all amenities from the Overpass API response
+  fun filterAllAmenities(responseJson: String): List<Facility> {
+    val reader = JsonReader(responseJson.reader())
+    val facilities = mutableListOf<Facility>()
+
+    reader.beginObject() // We're in the root object
+    while (reader.hasNext()) {
+      when (reader.nextName()) {
+        "elements" -> { // Start reading the elements array
+          reader.beginArray()
+          while (reader.hasNext()) {
+            reader.beginObject() // Begin a new element
+            var lat: Double? = null
+            var lon: Double? = null
+            var amenity: String? = null
+
+            // Read the object
+            while (reader.hasNext()) {
+              when (reader.nextName()) {
+                "lat" -> lat = reader.nextDouble()
+                "lon" -> lon = reader.nextDouble()
+                "tags" -> {
+                  reader.beginObject() // Read tags object
+                  while (reader.hasNext()) {
+                    val tagName = reader.nextName()
+                    if (tagName == "amenity") {
+                      amenity = reader.nextString()
+                    } else {
+                      reader.skipValue() // Skip other tag values
+                    }
+                  }
+                  reader.endObject() // End tags object
+                }
+                else -> reader.skipValue() // Skip all other keys
+              }
+            }
+
+            if (amenity != null && lat != null && lon != null) {
+              facilities.add(Facility(amenity, GeoPoint(lat, lon)))
+            }
+
+            reader.endObject() // End the element object
+          }
+          reader.endArray() // End the elements array
+        }
+        else -> reader.skipValue() // Skip unknown fields
+      }
+    }
+    return facilities
+  }
+}
--- a/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
+++ b/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
@@ -14,17 +14,32 @@ import org.osmdroid.util.GeoPoint
 
 class FacilitiesRepository(private val client: OkHttpClient) {
 
+  /**
+   * Sets a standard header for API requests.
+   *
+   * @param request Request builder instance to which headers will be added
+   */
   private fun setRequestHeaders(request: Request.Builder) {
     request.header("User-Agent", "Hikemate/1.0")
   }
 
+  /**
+   * Asynchronously fetches facilities within the bounds. Makes a request to the Overpass API to
+   * retrieve amenities like toilets, parking areas, and waste baskets. These can be specified in
+   * the FacilityType enum.
+   *
+   * @param bounds Geographical bounds within which to search for facilities
+   * @param onSuccess Callback to handle the resulting list of facilities when the operation
+   *   succeeds
+   * @param onFailure Callback invoked when the operation fails
+   */
   fun getFacilities(
       bounds: Bounds,
       onSuccess: (List<Facility>) -> Unit,
       onFailure: (Exception) -> Unit
   ) {
 
-    // looks like: "toilets|parking|waste_basket...
+    // Generate pipe-separated list of amenities (e.g., "toilets|parking|waste_basket...")
     val listOfAmenities = FacilityType.listOfAmenitiesForOverpassRequest()
 
     val requestData =
@@ -35,7 +50,6 @@ class FacilitiesRepository(private val client: OkHttpClient) {
         );
         out geom;
     """
-            .trimIndent()
 
     val requestBuilder =
         Request.Builder().url("https://overpass-api.de/api/interpreter?data=$requestData").get()
@@ -68,7 +82,14 @@ class FacilitiesRepository(private val client: OkHttpClient) {
             })
   }
 
-  fun filterAmenities(jsonString: String): MutableList<Facility> {
+  /**
+   * Parses and filters JSON response from Overpass API into a list of Facility objects. Filters out
+   * amenities that have an invalid latitude, longitude or facility type.
+   *
+   * @param jsonString JSON response string from Overpass API
+   * @return List of parsed Facility objects
+   */
+  fun filterAmenities(jsonString: String): List<Facility> {
     val gson = Gson()
     val facilities = mutableListOf<Facility>()
 
@@ -81,7 +102,6 @@ class FacilitiesRepository(private val client: OkHttpClient) {
       val obj = element.asJsonObject
       val lat = obj["lat"]?.asDouble
       val lon = obj["lon"]?.asDouble
-
       val tags = obj.getAsJsonObject("tags")
       val amenity = tags?.get("amenity")?.asString?.let { FacilityType.fromString(it) }
 
--- a/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
+++ b/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
@@ -12,8 +12,6 @@ import okhttp3.Request
 import okhttp3.Response
 import org.osmdroid.util.GeoPoint
 
-data class Facility(val type: FacilityType, val coordinates: GeoPoint)
-
 class FacilitiesRepository(private val client: OkHttpClient) {
 
   private fun setRequestHeaders(request: Request.Builder) {
--- a/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
+++ b/app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
@@ -1,27 +1,9 @@
 package ch.hikemate.app.model.facilities
 
-import android.util.Log
 import ch.hikemate.app.model.route.Bounds
-import com.google.gson.Gson
-import com.google.gson.JsonObject
-import java.io.IOException
-import okhttp3.Call
-import okhttp3.Callback
-import okhttp3.OkHttpClient
-import okhttp3.Request
-import okhttp3.Response
-import org.osmdroid.util.GeoPoint
 
-class FacilitiesRepository(private val client: OkHttpClient) {
-
-  /**
-   * Sets a standard header for API requests.
-   *
-   * @param request Request builder instance to which headers will be added
-   */
-  private fun setRequestHeaders(request: Request.Builder) {
-    request.header("User-Agent", "Hikemate/1.0")
-  }
+/** Interface for the facilities provider repository. */
+fun interface FacilitiesRepository {
 
   /**
    * Asynchronously fetches facilities within the bounds. Makes a request to the Overpass API to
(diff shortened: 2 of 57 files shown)


app/build.gradle.kts
```diff
@@ -174,6 +174,9 @@ dependencies {
     implementation(libs.googleid)
     implementation(libs.accompanist.permissions)
     implementation(libs.play.services.location)
+    implementation(libs.gson)
+
+
```
rBergCode, line 179: Question: Is there a reason to use `gson` and not `android.util.JsonReader` (overpass API) or `kotlinx.serialization` (elevation API) ?
Nitpick: you added spaces here.

app/src/main/java/ch/hikemate/app/model/facilities/Facility.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.osmdroid.util.GeoPoint
+
+/**
+ * Represents a facility or amenity point. Each facility has a specific type (e.g., toilet, parking)
+ * and geographical coordinates.
+ *
+ * @property type The type of facility, defined in [FacilityType]
+ * @property coordinates Geographic coordinates of the facility
+ */
+data class Facility(val type: FacilityType, val coordinates: GeoPoint)
```
rBergCode, line 12: Question: Why would you not use `LatLong` as in the rest of the app ? This would be easier in order to have the same type everywhere.

app/src/main/java/ch/hikemate/app/model/facilities/Facility.kt
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return String containing all facility types separated by pipe characters
+     */
+    fun listOfAmenitiesForOverpassRequest(): String {
+      var string = ""
+      for (facility in FacilityType.values()) {
+        string += facility.type + "|"
+      }
+      return string
+    }
```
rBergCode, line 47: Praise: Nice! Very clean

app/src/main/java/ch/hikemate/app/model/facilities/Facility.kt
```diff
@@ -0,0 +1,72 @@
...
+        "waste_basket" -> WASTE_BASKET
+        "supermarket" -> SUPERMARKET
+        "drinking_water" -> DRINKING_WATER
+        "ranger_station" -> RANGER_STATION
+        "bbq" -> BBQ
+        "bench" -> BENCH
+        "restaurant" -> RESTAURANT
+        "biergarten" -> BIERGARTEN
+        else -> null
+      }
```
rBergCode, line 69: Suggestion: You could replace that with a `for loop`. This would be easier to maintain if you add more facilities.
```suggestion
for (facility in FacilityType.values()) {
  if (facility.type == string) {
    return facility
  }
}
return null
```

app/src/test/java/ch/hikemate/app/model/facilities/FacilityTest.kt
```diff
@@ -0,0 +1,21 @@
...
+
+class FacilityTest {
+
+  @Test
+  fun testListOfAmenitiesForOverpassRequest() {
+    val list = FacilityType.listOfAmenitiesForOverpassRequest()
+    for (facility in FacilityType.values()) {
+      assert(list.contains(facility.type))
+    }
+  }
```
rBergCode, line 13: Praise: Nice check!

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -36,6 +36,29 @@ data class Bounds(val minLat: Double, val minLon: Double, val maxLat: Double, va
...
+   * Formats the bounds as a comma-separated string for use in Overpass API queries. Returns
+   * coordinates in the format required by Overpass API: "south,west,north,east"
+   * (minLat,minLon,maxLat,maxLon)
+   *
+   * @return String representation of the bounds in Overpass API format
+   */
+  fun toStringForOverpassAPI(): String {
+    return "$minLat,$minLon,$maxLat,$maxLon"
+  }
+
```
rBergCode, line 61: Suggestion: Nice to have a function for this! Can you use it also in `HikeRoutesRepositoryOverpass` on line 51

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -36,6 +36,29 @@ data class Bounds(val minLat: Double, val minLon: Double, val maxLat: Double, va
...
+  /**
+   * Checks if a given coordinate point falls within these bounds. Uses inclusive range checks for
+   * both latitude and longitude.
+   *
+   * @param lat The latitude to check, in decimal degrees
+   * @param lon The longitude to check, in decimal degrees
+   * @return true if the coordinate is within the bounds, false otherwise
+   */
+  fun containsCoordinate(lat: Double, lon: Double): Boolean {
+    return (lat in minLat..maxLat) && (lon in minLon..maxLon)
```
rBergCode, line 48: Praise: Very clean!

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
```diff
@@ -0,0 +1,115 @@
...
+                    Log.d("FacilitiesRepository", "Got ${facilities.size} facilities")
+                    onSuccess(facilities)
+                  } else {
+                    throw Exception("Request failed with code: ${response.code}")
+                  }
+                } catch (e: Exception) {
+                  onFailure(e)
+                }
+              }
+            })
```
rBergCode, line 82: Praise: Clean code here. Nice job!

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * Parses and filters JSON response from Overpass API into a list of Facility objects. Filters out
+   * amenities that have an invalid latitude, longitude or facility type.
+   *
+   * @param jsonString JSON response string from Overpass API
+   * @return List of parsed Facility objects
+   */
+  fun filterAmenities(jsonString: String): List<Facility> {
```
rBergCode, line 92: I would suggest rename it to `parseAmenities` as it is more explicit of what it does (even if it filters them out)

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
```diff
@@ -0,0 +1,115 @@
...
+      onSuccess: (List<Facility>) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+
+    // Generate pipe-separated list of amenities (e.g., "toilets|parking|waste_basket...")
+    val listOfAmenities = FacilityType.listOfAmenitiesForOverpassRequest()
+
+    val requestData =
+        """
+        [out:json];
```
rBergCode, line 47: As done in the overpass API, you could put this in a constant.

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
```diff
@@ -0,0 +1,115 @@
...
+        (
+          node["amenity"~"$listOfAmenities"](${bounds.toStringForOverpassAPI()});        
+        );
+        out geom;
+    """
+
+    val requestBuilder =
+        Request.Builder().url("https://overpass-api.de/api/interpreter?data=$requestData").get()
```
rBergCode, line 55: As done in the overpass API, you could put this in a constant.

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
```diff
@@ -0,0 +1,115 @@
...
+class FacilitiesRepository(private val client: OkHttpClient) {
+
+  /**
+   * Sets a standard header for API requests.
+   *
+   * @param request Request builder instance to which headers will be added
+   */
+  private fun setRequestHeaders(request: Request.Builder) {
+    request.header("User-Agent", "Hikemate/1.0")
+  }
```
rBergCode, line 24: Warning: There is the exact same function in the overpass API. This joins the global comment on this file.

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesRepository.kt
rBergCode, line 1: Warning: This is very clean. I have 2 remarks:
- Is there a reason to not put this code in the overpass API directly ? This is directly linked to it and most of the constants/functions are copied here.
- If you decide to keep it as a second file, you should create an interface to it so that we can pass an interface to the view model. This is better if we later want to move to a different API. (look at how overpass API is done)

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
```diff
@@ -0,0 +1,334 @@
...
+        val realFacilitiesRepository = FacilitiesRepository(OkHttpClient())
+        realFacilitiesRepository.getFacilities(
+            testBoundsNormal,
+            { routes ->
+              assertNotEquals(routes.size, 0)
+              print(routes)
+            }) {
+              fail("Should not fail")
+            }
+      }
```
rBergCode, line 121: Praise: Nice to have a real test here!

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
```diff
@@ -0,0 +1,334 @@
...
+  fun testGetFacilities_integrationTest() =
+      runTest(timeout = 5.seconds) {
+
+        // Calls the real API, not a mocked one
+        val realFacilitiesRepository = FacilitiesRepository(OkHttpClient())
+        realFacilitiesRepository.getFacilities(
+            testBoundsNormal,
+            { routes ->
+              assertNotEquals(routes.size, 0)
+              print(routes)
```
rBergCode, line 117: Warning: You should not leave print in the code. Either remove it or change it to a log.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
rBergCode, line 121: Suggestion: Can you make sure that the callback is called using a variable

```
...
var successCalled = false
  assertNotEquals(routes.size, 0)
  successCalled = true
...
assert(successCalled)
```

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
```diff
@@ -0,0 +1,334 @@
...
+      }
+
+  @Test
+  fun testGetFacilities_callsClient() {
+    `when`(client.newCall(any())).thenReturn(mock())
+    facilitiesRepository.getFacilities(
+        testBoundsNormal,
+        { routes ->
+          assertNotEquals(routes.size, 0)
+          print(routes)
```
rBergCode, line 130: Warning: print instead of log

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
```diff
@@ -0,0 +1,334 @@
...
+                VALID_RESPONSE_FOOTER) // There are no nodes returned in this response
+
+    setupMockResponse(response)
+
+    facilitiesRepository.getFacilities(
+        bounds = testBoundsNormal,
+        onSuccess = { facilities -> assertEquals(0, facilities.size) },
+        onFailure = { _ -> fail("getFacilities call with empty response failed") })
+
+    verify(mockCall).enqueue(callbackCaptor.capture())
```
rBergCode, line 212: Warning: same remark as above. Please check the the callback is called with a boolean

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
```diff
@@ -0,0 +1,334 @@
...
+          assertEquals(1, facilities.size)
+          assertEquals(FacilityType.TOILETS, facilities[0].type)
+          latch.countDown()
+          Log.d(
+              "FacilitiesRepositoryTest",
+              "getFacilities with normal parameters succeeded. Got: ${facilities.size} facilities")
+        },
+        onFailure = { fail("getFacilities call failed") })
+
+    verify(mockCall).enqueue(callbackCaptor.capture())
```
rBergCode, line 160: Warning: same remark as above. Please check the the callback is called with a boolean

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
rBergCode, line 1: Praise: Very nice tests. Very complete and you test most of the possible errors. Nice job!

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+          // Filters out facilities that are not within the requested bounds
+          val filteredFacilities =
+              cacheItem.value.filter { facility ->
+                bounds.containsCoordinate(
+                    facility.coordinates.latitude, facility.coordinates.longitude)
+              }
+          onSuccess(filteredFacilities)
+          return@withContext
+        }
+      }
```
rBergCode, line 48: Praise: Nice way to use the cache. I could argue that you may have all the facilities you request if you get from 2 parts of the cache but that would make the algorithm take way more time.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    `when`(mockFacilitiesRepository.getFacilities(any(), any(), any())).then {
+      val onSuccess = it.getArgument<(List<Facility>) -> Unit>(1)
+      onSuccess(facilities)
+    }
+
+    facilitiesViewModel.getFacilities(
+        bounds = testBounds,
+        onSuccess = { result -> assertEquals(facilities, result) },
+        onFailure = { fail("Should not be called") })
```
rBergCode, line 62: Warning: Please check the the callback is called with a boolean

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    `when`(mockFacilitiesRepository.getFacilities(any(), any(), any())).then {
+      val onFailure = it.getArgument<(Exception) -> Unit>(2)
+      onFailure(exception)
+    }
+
+    facilitiesViewModel.getFacilities(
+        bounds = testBounds,
+        onSuccess = { fail("Should not be called") },
+        onFailure = { ex -> assertEquals(exception, ex) })
```
rBergCode, line 77: Warning: same remark as above. Please check the the callback is called with a boolean

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    // Second call to getFacilities should use the cache
+    facilitiesViewModel.getFacilities(
+        bounds = testBounds,
+        onSuccess = { result -> assertEquals(facilities, result) },
+        onFailure = { fail("Should not be called") })
+
+    testDispatcher.scheduler.advanceUntilIdle()
+
+    verify(mockFacilitiesRepository, times(1)).getFacilities(any(), any(), any())
+    assertEquals(1, onSuccessCallCount)
```
rBergCode, line 110: Warning: You should have 2 success calls (and therefore add an incrementation in the second call

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -0,0 +1,183 @@
...
+    // calls bounds
+    facilitiesViewModel.getFacilities(
+        bounds = Bounds(46.501, 6.601, 46.509, 6.619),
+        onSuccess = { result -> assertEquals(facilities, result) },
+        onFailure = { fail("Should not be called") })
+
+    testDispatcher.scheduler.advanceUntilIdle()
+
+    verify(mockFacilitiesRepository, times(1)).getFacilities(any(), any(), any())
+    assertEquals(1, onSuccessCallCount)
```
rBergCode, line 144: Warning: Same remark as above. This should be 2

rBergCode (changes requested):
Very nice job. The code is very clean and most of my remarks are small changes.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
davidholzwarth: done

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryTest.kt
davidholzwarth: done

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
davidholzwarth: done

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
davidholzwarth: You're absolutely right. I've added a separate variable for each call, to ensure each one is only called once.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
davidholzwarth: done

app/build.gradle.kts
rBergCode: Thanks for the explanation! I think it's fine to have it. Just wanted to know why.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -47,41 +48,80 @@ class FacilitiesViewModelTest {
...
+          },
+          onFailure = { fail("Should not be called") })
+
+      testDispatcher.scheduler.advanceUntilIdle()
+
+      assertTrue(onSuccessCalled)
+    } finally {
+      Dispatchers.resetMain()
+      assertTrue(onSuccessCalled)
+    }
```
rBergCode, line 82: Question: Why are you using a try..catch here ? I've not seen anthing like this in the tests yet.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
```diff
@@ -47,41 +48,80 @@ class FacilitiesViewModelTest {
...
+            onFailureCalled = true
+          })
+
+      testDispatcher.scheduler.advanceUntilIdle()
+
+      assertTrue(onFailureCalled)
+    } finally {
+      Dispatchers.resetMain()
+      assertTrue(onFailureCalled)
+    }
```
rBergCode, line 115: Same question as above

app/src/main/java/ch/hikemate/app/model/overpassAPI/OverpassRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+  protected open fun buildAndSendRequest(
+      requestData: String,
+      client: OkHttpClient,
+      callback: Callback
+  ) {
+    val requestBuilder = Request.Builder().url("$OVERPASS_API_URL?data=$requestData").get()
+
+    client.newCall(requestBuilder.build()).enqueue(callback)
+  }
+}
```
rBergCode, line 42: Warning: I really like this function. This make the code really clean in the actual implementation of the API. But I think you forgot to set the request's header with a call to `setRequestHeaders`.

rBergCode (changes requested):
Pretty big changes but good! It's way cleaner now that we use a common class for the APIs. I have small question and remarks before you can merge it.

app/src/main/java/ch/hikemate/app/model/overpassAPI/OverpassRepository.kt
davidholzwarth: good catch, fixed

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
davidholzwarth: Its `try...finally`. In order to not mess up other tests we need to clean up the Dispatchers. One way to do it is to add the Dispatcher setup to an `@Before` function and clean them up in an `@After` function, however since many tests don't use dispatchers that would be quite a bit of unnecessary boilerplate. I've decided that this method is more efficient. The finally clause always executes, even if there is an error within the test, so this way we guarantee that the dispatchers definitely get cleaned up before the next test starts.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
davidholzwarth: Its `try...finally`. In order to not mess up other tests we need to clean up the Dispatchers. One way to do it is to add the Dispatcher setup to an `@Before` function and clean them up in an `@After` function, however since many tests don't use dispatchers that would be quite a bit of unnecessary boilerplate. I've decided that this method is more efficient. The finally clause always executes, even if there is an error within the test, so this way we guarantee that the dispatchers definitely get cleaned up before the next test starts.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesViewModelTest.kt
rBergCode: I see. Seems fine to me!

rBergCode (approved):
Thanks for the fix!

PR #200, opened 2024-11-22
Fix/Sign up button not shown

Wylwi:
# Summary
- Sign up button is cropped in small screen, due to the absolute positioning of the Button
- This has enabled us to discover other problems, which are resolved in this PR, unrelated to the scope of this task, but necessary to pass the tests. See the below details

# Details
- I changed the absolute position behavior to a Column and positioning system (similar to HTML Flex box), to make it scalable on every phone size
- I changed the paddings for the insets to use an android made function to handle it correctly, and no longer manually
- I made several edits to the End2End too, like making the button display being asserted, in addition to pressing it, since pressing it doesn't assert it's displayed neither than it's clickable
- The End2End should be more reliable too, with an improvement in the sign-out process of the End2End

# Issue fixed
The main issue fixed in this PR (i.e. the sign-up button not being shown) was because the Edge2Edge behavior was not handled correctly. I hence fixed it, but putting the correct padding on screens that were having a supplementary padding for the status bar, with the correct function. 

This led to the E2E failing because of a button being pushed out of the screen in the create account screen. This bug was indeed a recurrent bug, we failed to identify before. The fix was hard to find, but quite simple, it's just to retract the keyboard before checking if the button is there or not. This took a loooong time to find it.

This leads to another issue, a UI one. With the keyboard, the text fields were getting crushed. This was solved by making the screen scrollable. I fixed it on the 2 other screens with TextField too (SignInWithEmail and EditProfile).

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
@@ -17,7 +17,7 @@ import ch.hikemate.app.ui.auth.CreateAccountScreen
 import ch.hikemate.app.ui.auth.SignInScreen
 import ch.hikemate.app.ui.auth.SignInWithEmailScreen
 import ch.hikemate.app.ui.components.CenteredLoadingAnimation
-import ch.hikemate.app.ui.map.MapScreen.TEST_TAG_MAP
+import ch.hikemate.app.ui.map.MapScreen
 import ch.hikemate.app.ui.navigation.LIST_TOP_LEVEL_DESTINATIONS
 import ch.hikemate.app.ui.navigation.Screen
 import ch.hikemate.app.ui.navigation.Screen.PROFILE
@@ -109,10 +109,11 @@ class EndToEndTest : TestCase() {
     composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_SIGN_UP_BUTTON).performClick()
 
     // Wait for the map to load
-    composeTestRule.waitUntilExactlyOneExists(hasTestTag(TEST_TAG_MAP), timeoutMillis = 10000)
+    composeTestRule.waitUntilExactlyOneExists(
+        hasTestTag(MapScreen.TEST_TAG_MAP), timeoutMillis = 10000)
 
     // Check that we are on the map
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsDisplayed()
 
     // Check that the menu is displayed
     composeTestRule.onNodeWithTag(TEST_TAG_BOTTOM_BAR).assertIsDisplayed()
@@ -132,7 +133,7 @@ class EndToEndTest : TestCase() {
     composeTestRule
         .onNodeWithTag(TEST_TAG_MENU_ITEM_PREFIX + TopLevelDestinations.SAVED_HIKES.route)
         .performClick()
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsNotDisplayed()
+    composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsNotDisplayed()
     composeTestRule
         .onNodeWithTag(SavedHikesScreen.TEST_TAG_SAVED_HIKES_SECTION_CONTAINER)
         .assertIsDisplayed()
@@ -141,7 +142,7 @@ class EndToEndTest : TestCase() {
     composeTestRule
         .onNodeWithTag(TEST_TAG_MENU_ITEM_PREFIX + TopLevelDestinations.MAP.route)
         .performClick()
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsDisplayed()
     composeTestRule
         .onNodeWithTag(SavedHikesScreen.TEST_TAG_SAVED_HIKES_SECTION_CONTAINER)
         .assertIsNotDisplayed()
@@ -150,7 +151,7 @@ class EndToEndTest : TestCase() {
     composeTestRule
         .onNodeWithTag(TEST_TAG_MENU_ITEM_PREFIX + TopLevelDestinations.PROFILE.route)
         .performClick()
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsNotDisplayed()
+    composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsNotDisplayed()
 
     composeTestRule.waitUntilDoesNotExist(
         hasTestTag(CenteredLoadingAnimation.TEST_TAG_CENTERED_LOADING_ANIMATION),
@@ -162,6 +163,6 @@ class EndToEndTest : TestCase() {
     composeTestRule
         .onNodeWithTag(TEST_TAG_MENU_ITEM_PREFIX + TopLevelDestinations.MAP.route)
         .performClick()
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(MapScreen.TEST_TAG_MAP).assertIsDisplayed()
   }
 }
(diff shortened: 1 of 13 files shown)


app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -114,11 +120,12 @@ fun CreateAccountScreen(
       modifier =
           Modifier.testTag(Screen.CREATE_ACCOUNT)
               .padding(
-                  // Add for the status bar
                   start = 16.dp,
                   end = 16.dp,
-                  top = 40.dp,
-              ),
+              )
+              .safeDrawingPadding()
```
MaelImhof, line 126: Praise: this is a simple addition, but the kind of simple that takes hours to find. Very nice!

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -114,11 +120,12 @@ fun CreateAccountScreen(
...
+              .verticalScroll(scrollState)
+              .imePadding(),
```
MaelImhof, line 128: Question: I observed that this works in practice when running the app, but I'm not yet sure how it works exactly. Wasn't the `android:windowSoftInputMode="adjustResize"` you added to `AndroidManifest.xml` sufficient?

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
```diff
@@ -131,23 +138,26 @@ fun CreateAccountScreen(
...
 
         OutlinedTextField(
             modifier = Modifier.fillMaxWidth().testTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT),
             keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Email),
             colors = inputColors,
             value = email,
             onValueChange = { email = it },
-            label = { Text(stringResource(R.string.create_account_email_label)) })
+            label = { Text(stringResource(R.string.create_account_email_label)) },
+            singleLine = true)
```
MaelImhof, line 151: Praise: nice addition, we should have about that before!

app/src/main/java/ch/hikemate/app/ui/auth/SignInWithEmailScreen.kt
```diff
@@ -81,11 +88,13 @@ fun SignInWithEmailScreen(
...
-                  // Add for the status bar
                   start = 16.dp,
                   end = 16.dp,
-                  top = 40.dp,
-              ),
+              )
+              .safeDrawingPadding()
+              .verticalScroll(scrollState)
+              .imePadding()
+              .height(IntrinsicSize.Max),
```
MaelImhof, line 97: Question: what does this do exactly?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -93,17 +94,17 @@ object HikeDetailScreen {
...
+  const val TEST_TAG_HIKE_NAME = "HikeDetailScreenHikeName"
+  const val TEST_TAG_BOOKMARK_ICON = "HikeDetailScreenBookmarkIcon"
+  const val TEST_TAG_ELEVATION_GRAPH = "HikeDetailScreenElevationGraph"
+  const val TEST_TAG_DETAIL_ROW_TAG = "HikeDetailScreenDetailRowTag"
+  const val TEST_TAG_DETAIL_ROW_VALUE = "HikeDetailScreenDetailRowValue"
+  const val TEST_TAG_ADD_DATE_BUTTON = "HikeDetailScreenAddDateButton"
+  const val TEST_TAG_PLANNED_DATE_TEXT_BOX = "HikeDetailScreenPlannedDateTextBox"
+  const val TEST_TAG_DATE_PICKER = "HikeDetailDatePicker"
+  const val TEST_TAG_DATE_PICKER_CANCEL_BUTTON = "HikeDetailDatePickerCancelButton"
+  const val TEST_TAG_DATE_PICKER_CONFIRM_BUTTON = "HikeDetailDatePickerConfirmButton"
```
MaelImhof, line 107: Praise: thank you!

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -99,11 +104,11 @@ fun EditProfileScreen(
...
             Modifier.testTag(Screen.EDIT_PROFILE)
                 .padding(
-                    // Add for the status bar
                     start = 16.dp,
                     end = 16.dp,
-                    top = 16.dp,
-                ),
+                )
+                .safeDrawingPadding()
+                .verticalScroll(scrollState),
```
MaelImhof, line 111: Question: should you use `.imePadding()` here too? I'm not sure I understand how it works.

MaelImhof (commented):
A lot of different bits and fixes in this PR, it makes me feel like the app is significantly more stable, and the tests too. Just a few questions because I don't know a lot about padding, but other than that thank you for the various fixes you did.

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
Wylwi: According to [Android documentation](https://developer.android.com/develop/ui/compose/layouts/insets#insets-setup): Set android:windowSoftInputMode="adjustResize" in your Activity's AndroidManifest.xml entry. This setting **allows your app to receive the size of the software IME as insets**, which you can use to pad and lay out content appropriately when the IME appears and disappears in your app.

Thus, we receive the padding, but we still need to apply it!

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
Wylwi: Yep! I added the `imeNestedScroll` here too!

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -38,7 +39,7 @@ import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 class EndToEndTest : TestCase() {
-  @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
+  @get:Rule val composeTestRule = createComposeRule()
```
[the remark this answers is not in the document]
Wylwi: My issue is that I wanted to initialize the MainActivity when I want (especially, after the sign out). This was not possible with `createAndroidComposeRule<MainActivity>()`. My code using `HikeMateApp` was working but not really clean, I changed it with a more complex but clean version of it!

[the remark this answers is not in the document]
MaelImhof: I understand, thank you for your changes 👌🏻

app/src/androidTest/java/ch/hikemate/app/endtoend/EndToEndTest.kt
```diff
@@ -89,30 +93,59 @@ class EndToEndTest : TestCase() {
...
-    composeTestRule.onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT).performTextInput(email)
+
+    Espresso.closeSoftKeyboard()
+
+    composeTestRule
+        .onNodeWithTag(CreateAccountScreen.TEST_TAG_EMAIL_INPUT)
+        .assertIsDisplayed()
+        .performTextInput(email)
+
+    Espresso.closeSoftKeyboard()
```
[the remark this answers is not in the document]
MaelImhof: Yes that's what I thought, but we can't do the impossible.

app/src/main/java/ch/hikemate/app/ui/auth/CreateAccountScreen.kt
MaelImhof: Alright, makes no sense, as always with Android. But you made it work somehow anyway

MaelImhof: Alright, makes no sense to do things this way, as usual with Android, but I get it now.

MaelImhof (approved):
Thank you for your changes and (mostly) explanations. The code looks perfect to me now, ready to be merged 👌🏻

PR #201, opened 2024-11-22
feat: create a function that maps user location to location on hike

mchac15:
## Problem Statement
Need to accurately track user's progress along a hiking route, handling cases where the user's GPS location may not be exactly on the route path.

## Solution
Implemented a projection system that projects user location onto the nearest route segment and maintains route and segment information for path following

## Changes

- Added `segments` value to get route segments
- `RouteSegment`: Represents a segment of the route with start/end points and length
- `RouteProjectionResponse`: Contains projection results including:
  - Projected location
  - Current segment
  - Progress distance
  - Distance from route
  - Segment index

#### Core Functions
1. `projectLocationOnStart`:
   - Initial projection when starting route tracking
   - Uses greedy approach to find closest segment
   - Calculates initial progress distance

## In-Depth Algorithm Explanation

1. `projectLocationOnStart`:
The function uses a greedy approach to find the best initial projection:
- Validates input route has at least 2 points
- Special handling for single-segment routes
- For multi-segment routes:
  - Iterates through all segments
  - For each segment:
    - Projects point onto segment using vector projection
    - Calculates perpendicular distance to projection
    - If distance is smallest so far:
      - Updates closest segment
      - Updates progress distance based on:
        - Distance covered by previous segments
        - Distance from segment start to projection
  - Returns best projection found

- Selection:
   - Compares current and recursive projections
   - Selects projection with smallest distance from route
   - Returns best projection found

### Testing
Testing was difficult especially to find the projections of long hikes, I did my best to do the longest possible tests.

### Discussion
The previous `projectOnGoing` function was finally removed since it was unnecessary and was incorrectly accounting for the movement of the user. The PR contains a benchmark now which was used to determine the efficiency of the first algorithm. After some discussions it was concluded that we should keep only the first algorithm.


app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
MaelImhof, line 1: Nitpick: for consistency across the project, consider naming your tests using `pascalCase`, not `snake_case`. If you want them to be readable, you can also put their name in ` characters :

```
@Test
fun `My test name with spaces`() {
    // ...
}
```

(Yes this is cursed, but I use it a lot lol)

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
```diff
@@ -82,4 +82,120 @@ class HikeRouteTest {
...
+    val start = LatLong(60.0, 0.0)
+    val end = LatLong(60.0, 10.0)
+    val point = LatLong(61.0, 5.0)
+
+    val result = point.projectPointIntoLine(start, end)
+
+    // The projected point should be at lat=60.0 (on the line) and lon=5.0 (halfway)
+    // but the exact position should account for the latitude scaling
+    assertLatLongEquals(LatLong(60.0, 5.0), result)
+  }
```
MaelImhof, line 158: Question: I'm not sure what specific case this test is meant for. As far as I can tell, this is a horizontal line, just like in one of the previous tests. Plus, isn't it longitude scaling instead of latitude scaling? I probably understood wrong, those location things break my brain

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
```diff
@@ -82,4 +82,120 @@ class HikeRouteTest {
...
+  fun test_projection_with_exact_point_on_line() {
+    val start = LatLong(0.0, 0.0)
+    val end = LatLong(10.0, 10.0)
+    val point = LatLong(5.0, 5.0)
+
+    val result = point.projectPointIntoLine(start, end)
+
+    // Point should project to itself since it's already on the line
+    assertLatLongEquals(point, result)
+  }
```
MaelImhof, line 170: Praise: very good test case! Wouldn't have thought about it myself, but very important, well done!

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
```diff
@@ -82,4 +82,120 @@ class HikeRouteTest {
...
+
+    // Near poles, longitude differences should be heavily compressed
+    assertEquals(85.0, result.lat, 0.0001)
+    assertEquals(5.0, result.lon, 0.0001)
+  }
+
+  private fun assertLatLongEquals(expected: LatLong, actual: LatLong, delta: Double = 0.0001) {
+    assertEquals("Latitude should match", expected.lat, actual.lat, delta)
+    assertEquals("Longitude should match", expected.lon, actual.lon, delta)
+  }
```
MaelImhof, line 200: Praise: very clever ! Maybe just add the expected and actual values to the message so it's easier to debug.

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
MaelImhof, line 1: Suggestion: I did not see any test for a diagonal line with a point in the middle. This would help testing for the longitude adaptation based on latitude change, wouldn't it? For now, all the tests simply divide both coordinates differences by two and expect that as a result.

Maybe that was the purpose of `test_projection_accounts_for_latitude_scaling` and I did not understand it?

MaelImhof (commented):
You've done a lot of good job, well-documented and well thought-of code. I just have one point I want to discuss before reviewing your projection algorithm, please refer to my comments. Other than that, the code is very good, well done !

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -156,10 +186,43 @@ data class HikeRoute(
...
 ) {
+
   /** Get the color of the route from its id. The color should be the same for the same route id. */
   fun getColor(): Int {
     return hikeColors[abs(id.hashCode()) % hikeColors.size]
   }
+
+  /** The list of segments of the Route */
+  val segments: List<RouteSegment> by lazy {
+    return@lazy this.ways.windowed(2).map { (p1, p2) -> RouteSegment(p1, p2, p1.distanceTo(p2)) }
```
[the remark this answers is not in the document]
mchac15: To be completely honest I just didn't know this function existed, but that does make it easier.

app/src/main/java/ch/hikemate/app/utils/LocationUtils.kt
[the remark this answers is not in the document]
mchac15: Yes, your remark is honestly quite impressive I didn't even think of this case.
During the meeting on Monday it was discussed that the best way to handle this problem was to make two functions one for the initial projection, where the user just began to run the hike, and another function that was more efficient which calculated if the user advanced or not and update accordingly to save on performance.
I did not test the performance on the first one thinking the second one would solve that problem. If you know of a way how we could know if the user is advancing through the hike I can modify it. I thought of doing some kind of vector product between the difference of the traveled coordinates and the difference between the segment coordinates, but this still is wrong and I haven't found a way to account for this case.

What I am going to do:
1. If you know of a way to know if a user is advancing through the route I'll implement it.
2. If not I can remove this function entirely and add a threshold parameter to the other like for the second function

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
mchac15: The test is checking if our projection algorithm properly handles the fact that longitude degrees represent different distances at different latitudes (degrees get "compressed" as you move toward the poles)
I changed the test so it's clearer

app/src/main/java/ch/hikemate/app/utils/LocationUtils.kt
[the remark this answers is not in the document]
MaelImhof: Unfortunately I had an impediment on Monday, I couldn't make it to the meeting, however I'm not sure why we would need such optimization in the first place.

Because the logic is going to be run in the background every five seconds or so, there is plenty of time for it to terminate and for the UI to update I believe.

Maybe what you could do is :

- Create a test for the full projection (greedy approach)
   - Find a way to generate a hike with a lot of segments (for example, take the line of the equator and divide it into `n` pieces)
   - Create a large test hike with say 5000 or 10000 segments (which means double the number of waypoints, I think this is far more than what we have ever seen in terms of hike)
   - Create a simple test that measures the time it takes for the algorithm to terminate on this hike
   - If the time seems acceptable, we might as well simply use a non-optimized version in my book

What do you think of this?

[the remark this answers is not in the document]
mchac15: After some discussion on telegram I added a benchmark to the first algorithm to test its performance. Its performance was pretty good achieving 15 ms for 10 thousand points hike routes. 
I added the benchmark in the `adroidTest` folder

app/src/test/java/ch/hikemate/app/model/route/HikeRouteTest.kt
mchac15: I changed this test. You are right.

app/src/androidTest/java/ch/hikemate/app/benchmark/LocationProjectionBenchmark.kt
MaelImhof, line 1: Praise+Change: your benchmark is nicely implemented and helps us justify our discussion to only use full projection instead of using on-going updates as originally planned, congratulations!

However, those tests seem to be run by the CI as well, and they fail because of `ACTIVITY-MISSING CODE-COVERAGE DEBUGGABLE`. Maybe the simplest way to go would be to get the CI to ignore this class or those tests. I think it's possible but I don't know exactly.

It would be nice to have the tests be as simple as clicking the run button, but someone who pulls the repo will need to edit the Gradle config to run them correctly. I think this is out-of-scope for this task, unless you have a quick solution you want to try. Otherwise, we can address this later (or never), since this is an internal tool and not facing end-users.

app/src/test/java/ch/hikemate/app/utils/LocationUtilsTest.kt
MaelImhof, line 1: Suggestion: great testing! Maybe a test where there are more than one segment could be useful, to check that the segment selection is correct? No need for a million segment, maybe two or three would be enough.

MaelImhof (changes requested):
Thank you for your hard work on that benchmarking! I think this full projection approach will be way simpler to use in the *Run Hike* screen.

We are getting close, we'll merge this today

app/src/main/java/ch/hikemate/app/utils/LocationUtils.kt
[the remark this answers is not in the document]
MaelImhof: Great job on that benchmarking!

app/src/androidTest/java/ch/hikemate/app/benchmark/LocationProjectionBenchmark.kt
mchac15: I think we can just comment for now the tags @Test and  @Before

app/src/test/java/ch/hikemate/app/utils/LocationUtilsTest.kt
mchac15: Fixed.

app/src/androidTest/java/ch/hikemate/app/benchmark/LocationProjectionBenchmark.kt
MaelImhof: Yes, doing better might be out of scope for this task.

MaelImhof (approved):
Thank you for your changes and your amazing work! As far as I'm concerned, this PR is ready to be merged as soon as the CI passes.

PR #203, opened 2024-11-23
fix: date line issues for hikes crossing and searching

rBergCode:
# Summary
There was 2 issues about the date line:
1. Searching when showing the date line would make the app crash. For this, I changed it to make 2 requests, one on the left side and one on the right side.
2. Displaying a hike crossing the date line would make the app crash. For this, I filtered out the hikes doing this in the API.

# Details
I've not found any hikes in such conditions. This is prevention only.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
+++ b/app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
@@ -1,0 +1,19 @@
+package ch.hikemate.app.model.extensions
+
+import org.osmdroid.util.BoundingBox
+
+/** Return true if the bounding box crosses the date line */
+fun BoundingBox.crossesDateLine(): Boolean {
+  return lonEast < lonWest
+}
+
+/** Split the bounding box by the date line */
+fun BoundingBox.splitByDateLine(): Pair<BoundingBox, BoundingBox> {
+  return if (crossesDateLine()) {
+    Pair(
+        BoundingBox(latNorth, 180.0, latSouth, lonWest),
+        BoundingBox(latNorth, lonEast, latSouth, -180.0))
+  } else {
+    Pair(this, this)
+  }
+}
--- a/app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
+++ b/app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
@@ -1,5 +1,7 @@
 package ch.hikemate.app.model.extensions
 
+import android.util.Log
+import ch.hikemate.app.model.route.Bounds
 import org.osmdroid.util.BoundingBox
 
 /** Return true if the bounding box crosses the date line */
@@ -16,4 +18,11 @@ fun BoundingBox.splitByDateLine(): Pair<BoundingBox, BoundingBox> {
   } else {
     Pair(this, this)
   }
+}
+
+fun BoundingBox.toBounds(): Bounds {
+  Log.i(
+      "BoundingBox",
+      "Converting BoundingBox to Bounds(n:$latNorth, e:$lonEast, s:$latSouth, w:$lonWest)")
+  return Bounds(latSouth, lonWest, latNorth, lonEast)
 }
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
@@ -1,6 +1,5 @@
 package ch.hikemate.app.model.route
 
-import android.util.Log
 import ch.hikemate.app.ui.theme.hikeColors
 import kotlin.math.abs
 import kotlin.math.asin
@@ -77,13 +76,6 @@ data class Bounds(val minLat: Double, val minLon: Double, val maxLat: Double, va
       return Bounds(minLat, minLon, maxLat, maxLon)
     }
   }
-}
-
-fun BoundingBox.toBounds(): Bounds {
-  Log.i(
-      "BoundingBox",
-      "Converting BoundingBox to Bounds(n:$latNorth, e:$lonEast, s:$latSouth, w:$lonWest)")
-  return Bounds(latSouth, lonWest, latNorth, lonEast)
 }
 
 fun Bounds.toBoundingBox(): BoundingBox {
--- a/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
@@ -1,5 +1,6 @@
 package ch.hikemate.app.model.route
 
+import android.util.Log
 import ch.hikemate.app.ui.theme.hikeColors
 import kotlin.math.abs
 import kotlin.math.asin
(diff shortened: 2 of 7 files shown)


app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
```diff
@@ -51,6 +51,26 @@ open class ListOfHikeRoutesViewModel(
   private suspend fun getRoutesAsync(onSuccess: () -> Unit = {}, onFailure: () -> Unit = {}) {
     withContext(dispatcher) {
       val area = area_.value ?: return@withContext
+
+      // Check if the area is on the date line
+      if (area.lonEast < area.lonWest) {
```
MaelImhof, line 56: Suggestion: create an extension method on `BoundingBox` as you did with `Bounds` to make this condition easier to read.

```kotlin
fun BoundingBox.isCrossingDateLine(): Boolean {
    return lonEast < lonWest
}
```

app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
```diff
@@ -51,6 +51,26 @@ open class ListOfHikeRoutesViewModel(
...
+        val bounds1 = BoundingBox(area.latNorth, 180.0, area.latSouth, area.lonWest)
+        val bounds2 = BoundingBox(area.latNorth, area.lonEast, area.latSouth, -180.0)
```
MaelImhof, line 58: Suggestion: move this logic to an extension function as well.

```kotlin
fun BoundingBox.splitByDateLine(): Pair<BoundingBox, BoundingBox> {
    return if (isCrossingDateLine()) {
        Pair(
            BoundingBox(latNorth, 180.0, latSouth, lonWest),
            BoundingBox(latNorth, lonEast, latSouth, -180.0)
        )
    }
    else {
       Pair(this, this)
    }
}
```

Then, use it like in Scala (very satisfying) :

```kotlin
val (bounds1, bounds2) = area.splitByDateLine()
```

app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
```diff
@@ -51,6 +51,26 @@ open class ListOfHikeRoutesViewModel(
...
+            onSuccess = { routes1 ->
+              hikeRoutesRepository.getRoutes(
+                  bounds = bounds2.toBounds(),
+                  onSuccess = { routes2 ->
+                    hikeRoutes_.value = routes1 + routes2
+                    onSuccess()
+                  },
+                  onFailure = { _ -> onFailure() })
+            },
+            onFailure = { _ -> onFailure() })
```
MaelImhof, line 70: Question: is there any reason you did not do those two requests in parallel instead of sequentially? We might gain some time there, since both request will probably take time.

app/src/test/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpassTest.kt
```diff
@@ -1279,6 +1330,23 @@ class HikeRoutesRepositoryOverpassTest {
...
+    val callbackCapture = argumentCaptor<okhttp3.Callback>()
+
+    `when`(mockCall.enqueue(callbackCapture.capture())).then {
+      callbackCapture.firstValue.onResponse(mockCall, responseCrossingTheDateLine)
+    }
+
+    hikingRouteProviderRepositoryOverpass.getRoutes(
+        bounds, { routes -> assertEquals(0, routes.size) }) {
+          fail("Failed to fetch routes from Overpass API")
+        }
```
MaelImhof, line 1347: Suggestion: make sure that one of the callbacks is called.

```suggestion
    var successCalled = false
    hikingRouteProviderRepositoryOverpass.getRoutes(
        bounds, { routes -> assertEquals(0, routes.size); successCalled = true }) {
          fail("Failed to fetch routes from Overpass API")
        }
        
    assertTrue(successCalled)
```

MaelImhof (changes requested):
Very nice work on this fix ! A few refactoring concerns and testing nitpicks, and the PR will be ready 👌🏻

app/src/main/java/ch/hikemate/app/model/route/HikeRoute.kt
```diff
@@ -68,13 +69,18 @@ data class Bounds(val minLat: Double, val minLon: Double, val maxLat: Double, va
...
+    }
+
     fun build(): Bounds {
       return Bounds(minLat, minLon, maxLat, maxLon)
     }
   }
 }
 
 fun BoundingBox.toBounds(): Bounds {
+  Log.i("BoundingBox", " searching at n:$latNorth, e:$lonEast, s:$latSouth, w:$lonWest")
```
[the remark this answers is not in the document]
rBergCode: You're right!

app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
rBergCode: Nice!

app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
rBergCode: nice!

app/src/test/java/ch/hikemate/app/model/route/HikeRoutesRepositoryOverpassTest.kt
rBergCode: Nice catch!

app/src/main/java/ch/hikemate/app/model/route/ListOfHikeRoutesViewModel.kt
rBergCode: How would you manage the callbacks in this case ? I'm not sure how to do it.

MaelImhof: Fair, it would make the code a little bit more complex and I can't tell you how to do it from the top of my head now, so I guess we can call it a day and fix this later if we notice strong problems with this additional delay.

app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
MaelImhof, line 1: Nitpick: either `BoundingBox.toBounds()` should be moved to this file, or those extensions should be moved to `HikeRoute.kt`, but it's not coherent to have them separate.

MaelImhof (changes requested):
Thank you for your changes! One last nitpick and you're good to merge.

app/src/main/java/ch/hikemate/app/model/extensions/BoundingBoxExtension.kt
rBergCode: You're right. Thanks for the remark!

MaelImhof (approved):
Thanks for the nitpick changes, looks perfect to me now !

PR #205, opened 2024-11-24
ci: add logcat output as artifact in CI

Wylwi:
# Summary
- The objective of this PR is to help with debugging issues that happen only in CI, by providing logcat outputs in the artifacts.

# Details
- One difficulty is to ensure that the exit code provided by the tests is not lost during the process, that's why we store it, run logcat, and rethrow it


MaelImhof (approved):
Neat change to have the logcat to debug CI fails!

LGTM

PR #207, opened 2024-11-25
refactor: use AsyncStateHandler in profile

rBergCode:
# Summary
I've created an AsyncStateHandler component (#193) in order to reduce repeated code. In this PR, I will use it in all the places the code was repeated.

# Details
I also moved the getter for the display of a hiking level to reduce complexity pointed out by sonar.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
+++ b/app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
@@ -1,12 +1,21 @@
 package ch.hikemate.app.model.profile
 
+import ch.hikemate.app.R
 import com.google.firebase.Timestamp
 
 /** An enum class representing the hiking level of a user. */
 enum class HikingLevel {
   BEGINNER,
   INTERMEDIATE,
-  EXPERT
+  EXPERT;
+
+  fun getDisplayNameId(): Int {
+    return when (this) {
+      BEGINNER -> R.string.hiking_level_beginner
+      INTERMEDIATE -> R.string.hiking_level_intermediate
+      EXPERT -> R.string.hiking_level_expert
+    }
+  }
 }
 
 /**
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -112,9 +112,9 @@
 
     <!-- Edit Profile Screen -->
     <string name="edit_profile_screen_title">Edit your profile</string>
-    <string name="profile_screen_hiking_level_choice_beginner">Beginner</string>
-    <string name="profile_screen_hiking_level_choice_intermediate">Intermediate</string>
-    <string name="profile_screen_hiking_level_choice_expert">Expert</string>
+    <string name="hiking_level_beginner">Beginner</string>
+    <string name="hiking_level_intermediate">Intermediate</string>
+    <string name="hiking_level_expert">Expert</string>
     <string name="edit_profile_screen_save_button_text">Save profile</string>
 
     <!-- Authentication -->
--- a/app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
@@ -1,13 +1,10 @@
 package ch.hikemate.app.ui.profile
 
-import android.util.Log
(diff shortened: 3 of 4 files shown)


rBergCode:
The sonar complain is about reuse of code but this was kind of the point of this PR. Use the common `AsyncStateHandler` wherever it's useful.

app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
```diff
@@ -1,12 +1,21 @@
...
   INTERMEDIATE,
-  EXPERT
+  EXPERT;
+
+  fun getDisplayNameId(): Int {
+    return when (this) {
+      BEGINNER -> R.string.hiking_level_beginner
+      INTERMEDIATE -> R.string.hiking_level_intermediate
+      EXPERT -> R.string.hiking_level_expert
+    }
```
davidholzwarth, line 17: Suggestion: Nice, this is a way better placement for this function. 

However, I don't think the way the function currently needs to be used (first fetching id and then fetching the string that belongs to this id) is very intuitive. I've added a suggestion, which takes the context and returns a string directly, thus avoiding the detour with the StringID. If you accept the suggestion, make sure to also update the usages of this function in the code.

I'll leave it up to you to see which option you prefer, since it works the way it currently is too, but either way, please make sure to add a docstring to document this functions use, so that we can keep our code nice and understandable and maintainable.

```suggestion
  fun getDisplayString(context: Context): String {
    return when (this) {
      BEGINNER -> context.getString(R.string.hiking_level_beginner)
      INTERMEDIATE -> context.getString(R.string.hiking_level_intermediate)
      EXPERT -> context.getString(R.string.hiking_level_expert)
    }
```

app/src/main/res/values/strings.xml
```diff
@@ -112,9 +112,9 @@
 
     <!-- Edit Profile Screen -->
     <string name="edit_profile_screen_title">Edit your profile</string>
-    <string name="profile_screen_hiking_level_choice_beginner">Beginner</string>
-    <string name="profile_screen_hiking_level_choice_intermediate">Intermediate</string>
-    <string name="profile_screen_hiking_level_choice_expert">Expert</string>
+    <string name="hiking_level_beginner">Beginner</string>
+    <string name="hiking_level_intermediate">Intermediate</string>
+    <string name="hiking_level_expert">Expert</string>
```
davidholzwarth, line 117: Question: I like the renaming, but do you think it makes sense to add them to their own section, like "Profile" or something like that, so that its more intuitive that they can also be used outside of the profile screen in the future?

davidholzwarth (commented):
Nice PR, its always good to see the codebase cleaned up a little and bad practices being undone. :) Did I understand correctly, that the big changes in the ProfileScreen and EditProfileScreen are just you having added the AsyncStateHandler around the code? Its hard to tell, since GitHub marks it all as one big change, so maybe I overlooked some smaller changes within. 

The changes I did notice was the Hiking Expertise  Level toString function you mentioned in the PR, thats good, but I proposed a suggestion.

app/src/main/res/values/strings.xml
rBergCode: You're right. I changed that.

app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
rBergCode: Yes, I agree with you. I changed it

rBergCode:
About the issue with `End2EndTest2`, I also find it unreliable. This is something to be fixed and not related to this PR.

app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
davidholzwarth: Thanks. Could you please add a docstring as well? I am sorry for being so nitpicky, but thats a feedback we got from the coaches.

davidholzwarth (commented):
Thanks. Just a docstring missing, then it should be gtg.

app/src/main/java/ch/hikemate/app/model/profile/Profile.kt
rBergCode: Sure no worries!

davidholzwarth (approved):
Thanks!

PR #211, opened 2024-11-26
fix: change date picker to compose one

rBergCode:
# Summary
The issue was that the calendar had ugly colors. We needed to use the primary color.

# Details
As I investigated the issue, I found out we used the wrong `DatePickerDialog`. I changed it to use the compose dialog one.
Also, I updated the them in order to use the primary color app-wide. This changes most of the colors to use the correct one.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
@@ -1,7 +1,6 @@
 package ch.hikemate.app.ui.map
 
 import android.annotation.SuppressLint
-import android.app.DatePickerDialog
 import android.util.Log
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.Image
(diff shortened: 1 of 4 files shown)


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -426,7 +443,7 @@ fun DateDetailRow(
                   // saved Date
                   style = MaterialTheme.typography.bodySmall,
                   modifier =
-                      Modifier.clickable { showDatePickerDialog() }
+                      Modifier.clickable { showDatePicker() }
```
davidholzwarth, line 446: Request: LGMT, but I noticed that the border of the Box in which this date is is black. Could you change it to onPrimary please or something similar?

app/src/main/res/values/strings.xml
```diff
@@ -63,6 +63,10 @@
     <string name="hike_detail_screen_value_planned">Planned</string>
     <string name="hike_detail_screen_add_a_date_button_text">Add a date</string>
 
+    <!-- Hike Details Screen - Date Picker -->
+    <string name="hike_detail_screen_date_picker_cancel_button">Cancel</string>
+    <string name="hike_detail_screen_date_picker_confirm_button">Plan my hike!</string>
```
davidholzwarth, line 68: Nitpick: Maybe `Plan this hike!` sounds better? Or just `Plan this hike`?

davidholzwarth (commented):
Nice, looks way better. Two little nitpicks but otherwise you're good. :)

app/src/main/res/values/strings.xml
rBergCode: "Plan this hike!" sounds good.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
davidholzwarth: Thanks

davidholzwarth (approved):
LGTM now!

PR #218, opened 2024-11-27
refactor: replace and remove toFormattedDate

MaelImhof:
This PR removes `DateTimeUtils.toFormattedDate` and replaces it with `DateTimeUtils.humanReadableFormat`.

`DateTimeUtils.toFormattedDate` used a date format that did not adapt to the current Locale. It used the `dd/MM/yyyy` format, but this format can be mistaken for `MM/dd/yyyy` depending on the date, which can be confusing for English speakers who are used to the second option rather than the first.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
+++ b/app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
@@ -3,7 +3,6 @@ package ch.hikemate.app.utils
 import android.content.Context
 import ch.hikemate.app.R
 import com.google.firebase.Timestamp
-import java.text.SimpleDateFormat
 import java.time.Duration
 import java.time.LocalDate
 import java.time.LocalDateTime
@@ -83,9 +82,4 @@ fun Timestamp.humanReadablePlannedLabel(
     // Later than next year, for example "Planned on 18th of October 2026"
     else -> context.getString(R.string.datetime_utils_planned_on_date, formattedDate)
   }
-}
-
-fun Timestamp.toFormattedString(locale: Locale = Locale.getDefault()): String {
-  val dateFormat = SimpleDateFormat("dd/MM/yyyy", locale)
-  return dateFormat.format(this.toDate())
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
```diff
@@ -84,8 +83,3 @@ fun Timestamp.humanReadablePlannedLabel(
     else -> context.getString(R.string.datetime_utils_planned_on_date, formattedDate)
   }
 }
-
-fun Timestamp.toFormattedString(locale: Locale = Locale.getDefault()): String {
```
davidholzwarth, line 88: Question: IK its not really part of this PR specifically, but does `toHumanReadableFormat` also work for locales like Chinese and stuff? Could you help me understand what is going on in `humanReadableFormat`, because for me it low-key looks the same as the `toFormattedString` function

davidholzwarth (commented):
Just a question

app/src/main/java/ch/hikemate/app/utils/DatetimeUtils.kt
MaelImhof: It is in fact the same as `toFormattedString`, the format being the main difference. `humanReadableFormat` uses `d MMMM yyyy` as format, which will produce dates such as (in English)

- 1 January 2001
- 17 April 2013

It spells out the complete name of the month. Of course, the format itself is not adapted to the locale, but the month is clearly spelled out instead of a number.

`toFormattedString` uses `dd/mm/yyyy`, which can easily be confused with `mm/dd/yyyy`, so I just thought it might be a good idea to use `humanReadableFormat` as well.

However I'm not sure how well it would work with Chinese locale, never tested it

davidholzwarth: ok, thats fine by me. Thanks!

davidholzwarth (approved):
LGTM fr this time

PR #219, opened 2024-11-28
feat(hikes): HikesViewModel factory

MaelImhof:
# Summary

#192 introduces a new view model but does not implement a factory to initialize it. This quick PR adds a factory to the view model to create it properly in `MainActivity` like the others.

## Details

Pretty standard stuff, I copied the factory logic and test from `ListOfHikeRoutesViewModel`, with the difference that it needed to mock `FirebaseFirestore` and `FirebaseAuth` since it uses it to instantiate the `SavedHikesRepository` used internally.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
@@ -2,13 +2,18 @@ package ch.hikemate.app.model.route
 
 import android.util.Log
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
 import ch.hikemate.app.model.elevation.ElevationService
+import ch.hikemate.app.model.elevation.ElevationServiceRepository
 import ch.hikemate.app.model.extensions.toBounds
 import ch.hikemate.app.model.route.saved.SavedHike
 import ch.hikemate.app.model.route.saved.SavedHikesRepository
+import ch.hikemate.app.model.route.saved.SavedHikesRepositoryFirestore
 import ch.hikemate.app.utils.RouteUtils
 import com.google.firebase.Timestamp
+import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.firestore.FirebaseFirestore
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -40,6 +46,20 @@ class HikesViewModel(
...
+    val Factory: ViewModelProvider.Factory =
+        object : ViewModelProvider.Factory {
+          @Suppress("UNCHECKED_CAST")
+          override fun <T : ViewModel> create(modelClass: Class<T>): T {
+            return HikesViewModel(
+                savedHikesRepo =
+                    SavedHikesRepositoryFirestore(
+                        FirebaseFirestore.getInstance(), FirebaseAuth.getInstance()),
+                osmHikesRepo = HikeRoutesRepositoryOverpass(OkHttpClient()),
+                elevationService = ElevationServiceRepository(OkHttpClient()))
```
Wylwi, line 58: Nitpick: Couldn't you use the same OkHttpClient?

app/src/test/java/ch/hikemate/app/model/route/HikesViewModelTest.kt
```diff
@@ -136,6 +139,24 @@ class HikesViewModelTest {
...
+    every { FirebaseFirestore.getInstance() } returns firebaseFirestore
+    val firebaseAuth: FirebaseAuth = mockk()
+    mockkStatic(FirebaseAuth::class)
+    every { FirebaseAuth.getInstance() } returns firebaseAuth
+
+    val factory = HikesViewModel.Factory
+    val viewModel = factory.create(HikesViewModel::class.java)
+    assertNotNull(viewModel)
+  }
+
```
Wylwi, line 159: Praise: Simple but efficient test!

Wylwi (commented):
I just have a small comment about the OkHttpClient, otherwise, it's good to go after it!

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: That's a good idea, done 👌🏻

Wylwi: Thanks!

Wylwi (approved):
Thanks for your changes! You can merge it now!

PR #221, opened 2024-11-28
refactor: use jsonReader in facilities repo

rBergCode:
# Summary
The facilities repository was using a dependency called `gson`. This works but when we retrieve a lot of data, this library won't handle it and would throw an out of memory error.

To manage this, I changed the facilities repository to use the JSON reader.

### DIFF
--- a/app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
+++ b/app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
@@ -64,7 +64,7 @@ class FacilitiesRepositoryOverpassTest {
         }
       }"""
 
-    private const val VALID_RESPONSE_ELEMENT_WITH_MULTIPLE_TAGS =
+    private const val VALID_RESPONSE_ELEMENT_2 =
         """
       {
         "type": "node",
@@ -167,7 +167,7 @@ class FacilitiesRepositoryOverpassTest {
             VALID_RESPONSE_HEADER +
                 VALID_RESPONSE_ELEMENT +
                 "," +
-                VALID_RESPONSE_ELEMENT_WITH_MULTIPLE_TAGS +
+                VALID_RESPONSE_ELEMENT_2 +
                 VALID_RESPONSE_FOOTER)
 
     val callbackCaptor = argumentCaptor<okhttp3.Callback>()
--- a/app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
+++ b/app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
@@ -63,6 +63,18 @@ class FacilitiesRepositoryOverpassTest {
           "amenity": "toilets"
         }
       }"""
+
+    private const val VALID_RESPONSE_ELEMENT_WITH_MULTIPLE_TAGS =
+        """
+      {
+        "type": "node",
+        "id": 313827518,
+        "lat": 50.9620,
+        "lon": 5.12740,
+        "tags": {
+          "amenity": "parking"
+        }
+      }"""
   }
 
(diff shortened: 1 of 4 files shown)


app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
```diff
@@ -110,34 +106,6 @@ class FacilitiesRepositoryOverpassTest {
...
-              },
-              onFailure = { error -> continuation.resumeWithException(error) })
-        }
-      }
-    }
-
-    assertTrue(onSuccessCalled)
-    assertNotEquals(0, routesSize)
-  }
-
```
Wylwi, line 140: Question: Why was this test removed?

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
Wylwi, line 1: Nitpick: I see no tests for testing multiple facilities being retrieved, could you add one please?

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
rBergCode: Good question: I removed it as it is not a unit test and therefore should not be here.

Wylwi: Wasn't there somewhere else you could've put it instead of removing it? Or do we assume that this is going to be tested by a larger E2E test later on?

rBergCode: I will be tested by E2E 3 I guess.

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
rBergCode: Sure!

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
Wylwi: Okay, that makes sense!

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
```diff
@@ -63,6 +63,18 @@ class FacilitiesRepositoryOverpassTest {
...
+        """
+      {
+        "type": "node",
+        "id": 313827518,
+        "lat": 50.9620,
+        "lon": 5.12740,
+        "tags": {
+          "amenity": "parking"
+        }
+      }"""
```
Wylwi, line 77: Nitpick: This doesn't look like the val has multiple tags, could you improve the variable name please?

Wylwi (commented):
Thanks for the new test! I've only got one last nitpick and you'll be able to merge!

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
Wylwi: Thanks!

app/src/test/java/ch/hikemate/app/model/facilities/FacilitiesRepositoryOverpassTest.kt
rBergCode: Oops

Wylwi (approved):
Thanks for the changes! LGTM now

PR #223, opened 2024-11-28
feat: implement run this hike button

davidholzwarth:
# Summary

I added a "Run this hike" button to the hike details screen, which lets the user trigger our run this hike feature. Instead of passing the `navigationActions` to the composable which contains the button (`HikeDetails`) to then trigger the navigation to the run hike screen, I've passed a `onRunThisHike` callback, since I believe this makes more sense, so the `HikeDetails` composable does not need to worry about how to implement onRunThisHike at all, and the parent, the `HikeDetailScreen` composable has full control. 

# Note
I've changed all test tags in the hike details screen to fit our new naming convention of `{SCREEN}_TEST_TAG_{COMPONENT}` convention. This is since I had to add a new test tag, and it felt a little bit cheeky to add a single one that is conform with our convention, but leave the rest as they are...

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
@@ -22,18 +22,18 @@ import ch.hikemate.app.model.route.ListOfHikeRoutesViewModel
 import ch.hikemate.app.model.route.saved.SavedHikesRepository
 import ch.hikemate.app.model.route.saved.SavedHikesViewModel
 import ch.hikemate.app.ui.components.BackButton.BACK_BUTTON_TEST_TAG
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_ADD_DATE_BUTTON
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_BOOKMARK_ICON
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_DATE_PICKER
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_DATE_PICKER_CANCEL_BUTTON
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_DATE_PICKER_CONFIRM_BUTTON
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_DETAIL_ROW_TAG
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_DETAIL_ROW_VALUE
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_ELEVATION_GRAPH
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_HIKE_NAME
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_MAP
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_PLANNED_DATE_TEXT_BOX
-import ch.hikemate.app.ui.map.HikeDetailScreen.HIKE_DETAILS_TEST_TAG_RUN_HIKE_BUTTON
+import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_ADD_DATE_BUTTON
+import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_BOOKMARK_ICON
+import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_DATE_PICKER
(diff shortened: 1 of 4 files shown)


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -93,17 +96,18 @@ object HikeDetailScreen {
...
+  const val HIKE_DETAILS_TEST_TAG_BOOKMARK_ICON = "HikeDetailScreenBookmarkIcon"
+  const val HIKE_DETAILS_TEST_TAG_ELEVATION_GRAPH = "HikeDetailScreenElevationGraph"
+  const val HIKE_DETAILS_TEST_TAG_DETAIL_ROW_TAG = "HikeDetailScreenDetailRowTag"
+  const val HIKE_DETAILS_TEST_TAG_DETAIL_ROW_VALUE = "HikeDetailScreenDetailRowValue"
+  const val HIKE_DETAILS_TEST_TAG_ADD_DATE_BUTTON = "HikeDetailScreenAddDateButton"
+  const val HIKE_DETAILS_TEST_TAG_PLANNED_DATE_TEXT_BOX = "HikeDetailScreenPlannedDateTextBox"
+  const val HIKE_DETAILS_TEST_TAG_DATE_PICKER = "HikeDetailDatePicker"
+  const val HIKE_DETAILS_TEST_TAG_DATE_PICKER_CANCEL_BUTTON = "HikeDetailDatePickerCancelButton"
+  const val HIKE_DETAILS_TEST_TAG_DATE_PICKER_CONFIRM_BUTTON = "HikeDetailDatePickerConfirmButton"
+  const val HIKE_DETAILS_TEST_TAG_RUN_HIKE_BUTTON = "HikeDetailRunHikeButton"
```
rBergCode, line 110: Question: Wasn't the convention more about the value of the tag and not the variable name of it ?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -227,18 +231,34 @@ fun HikeDetailScreen(
                   .padding(bottom = MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT + 8.dp))
 
       // Hike Details bottom sheet
-      HikeDetails(detailedRoute, savedHikesViewModel, elevationData, profile.hikingLevel)
+      HikeDetails(
+          detailedRoute = detailedRoute,
+          savedHikesViewModel = savedHikesViewModel,
+          elevationData = elevationData,
+          userHikingLevel = profile.hikingLevel,
+          onRunThisHike = { navigationActions.navigateTo(Screen.RUN_HIKE) },
```
rBergCode, line 239: Question: I'm not sure to understand why you pass this as a parameter instead of just setting the `onClick` of the button ?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -227,18 +231,34 @@ fun HikeDetailScreen(
...
 
+/**
+ * A composable that displays details about a hike in a bottom sheet.
+ *
+ * @param detailedRoute The route for which information is displayed
+ * @param savedHikesViewModel ViewModel managing saved hikes state
+ * @param elevationData List of elevation points for the elevation graph
+ * @param userHikingLevel The user's hiking experience level
+ * @param onRunThisHike Callback triggered when "Run This Hike" button is clicked
+ */
```
rBergCode, line 253: Praise: Nice catch!

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -412,14 +432,35 @@ class HikeDetailScreenTest {
...
+          { onRunThisHike() })
+    }
+
+    composeTestRule
+        .onNodeWithTag(HIKE_DETAILS_TEST_TAG_RUN_HIKE_BUTTON)
+        .assertIsDisplayed()
+        .performClick()
+
+    verify(onRunThisHike).invoke()
+  }
```
rBergCode, line 465: Suggestion: I think you pass the action as a parameter because of this test. I think it's okay to test that you call a `navigateTo` in here. It won't work because you don't have a navigation here but you can mock it. (check other tests)

rBergCode (changes requested):
Good job of refactoring! I have some questions about it.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
davidholzwarth: test

rBergCode: Thanks!

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
rBergCode: Yes okay. Seems fair. I forgot that you would have to pass the `navigationActions`. Good to me then!

rBergCode (approved):
Thanks for the explanation and the small fix!

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
rBergCode: Yep!

PR #227, opened 2024-11-30
fix: planned hikes also appear in the saved hikes screen

mchac15:
# Summary
Little fix for the `SavedHikes` screen: now a planned hike also appears in the `SavedHikes` screen. This is possible by simply changing the filtering of the `savedHike` in the `savedHikesViewModel`.


davidholzwarth (approved):
LGTM

PR #229, opened 2024-11-30
feat: add Location button to the elevation graph

mchac15:
# Summary:

This PR adds a location marker feature to the `ElevationGraph` component to display user progress through a hike. Key changes:

- Added optional `progressThroughHike` parameter (0.0-1.0) to show current position (normalized)
- Added marker size customization through style properties 

The marker provides visual feedback for users during hike navigation while maintaining the existing graph functionality for other use cases.

### Notes
- The location marker reuses the existing MapUtils marker icon to maintain UI consistency across the app.
- It was decided to add the `progressThroughHike` rather than taking the position of the user since the projection would have to be recalculated, the projection function retrurns the distance of the hike that has already been traveled, so the calculation is very simple to make and handle in the other screens and avoids recalculation.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
@@ -1,6 +1,8 @@
 package ch.hikemate.app.ui.components
 
 import android.content.Context
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.size
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.test.*
@@ -9,6 +11,9 @@ import androidx.compose.ui.unit.dp
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import ch.hikemate.app.R
+import ch.hikemate.app.utils.MapUtils
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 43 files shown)


app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
```diff
@@ -0,0 +1,92 @@
...
+      ElevationGraph(
+          elevations = elevationData,
+          modifier = Modifier.size(200.dp),
+          progressThroughHike = 0.5f,
+          styleProperties = ElevationGraphStyleProperties(locationMarkerSize = 24f))
+    }
+
+    composeTestRule.waitForIdle()
+
+    // Since we can't directly test Canvas drawing, we verify no error states are shown
```
davidholzwarth, line 64: Warning: I feel quite uncomfortable just not testing whether the marker exists. I found a function called .captureToImage().readPixels(), which gets a pixel array of the drawable, where you could then test if the marker is in the array of pixels. Could you please try if that works for the tests?

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    composeTestRule
+        .onNode(hasText(context.getString(R.string.elevation_graph_loading_label)))
+        .assertDoesNotExist()
+    composeTestRule
+        .onNode(hasText(context.getString(R.string.hike_card_no_data_label)))
+        .assertDoesNotExist()
+  }
+
+  @Test
+  fun elevationGraph_withLargeDataSet_handlesDataCorrectly() {
```
davidholzwarth, line 74: Warning: here too, this test doesn'T really test anything. Perhaps you could check if the color of the hike is present in the image or something like that.

davidholzwarth (commented):
Good code, lovely implementation, however I am a bit concerned about the tests...

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import ch.hikemate.app.R
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
```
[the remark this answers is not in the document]
mchac15: We didn't

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
```diff
@@ -152,6 +159,46 @@ fun ElevationGraph(
...
               style = Stroke(width = styleProperties.strokeWidth))
           clipPath(path, clipOp = ClipOp.Intersect) { drawRect(color = styleProperties.fillColor) }
+
+          // If the progressThroughHike is not null then display the location marker
+          progressThroughHike?.let { progress ->
+            // The X component of the marker
+            val markerX = width * progress
+
+            val pointIndex = (progress * (scaledData.size - 1)).toInt()
+            val pointProgress = (progress * (scaledData.size - 1)) % 1
```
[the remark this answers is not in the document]
mchac15: It calculates the fractional progress between two points.

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
mchac15: Thanks for commenting this I didn't know it existed, I added tests using this function. 
The thing was that since the person who did the `ElevationGraph` used a `Canvas` it's impossible to add testTags to a `Canvas` so it was difficult for me to find a way to test whether the Marker was drawn or not, especially since the graph didn't have any tests.

app/src/main/java/ch/hikemate/app/ui/components/ElevationGraph.kt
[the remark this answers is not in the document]
davidholzwarth: Perfect, thats super helpful!

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    composeTestRule.waitForIdle()
+
+    // Verify that neither loading nor no data messages are shown
+    composeTestRule
+        .onNode(hasText(context.getString(R.string.elevation_graph_loading_label)))
+        .assertDoesNotExist()
+    composeTestRule
+        .onNode(hasText(context.getString(R.string.hike_card_no_data_label)))
+        .assertDoesNotExist()
```
davidholzwarth, line 51: Improvement: Could you also, armed with your knowledge of these pixel tests, please add a test here that shows that the composable contains the hike color? That would be a neat addition that confirms that the hike is displayed in the card.

mchac15: Here it should not be displayed. But I still added the check

davidholzwarth: I'm afraid you've misunderstood, I meant to check wether the color of the hike is on the image, so whether the hike elevation graph is displayed properly.

davidholzwarth (commented):
Almost there. IMO there is one test missing, see the unresolved conversation.

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
davidholzwarth: Thanks!

app/src/androidTest/java/ch/hikemate/app/ui/components/ElevationGraphTest.kt
davidholzwarth: looks great, thanks!

davidholzwarth (approved):
LGTM now :)

PR #232, opened 2024-11-30
feat(hikes): new HikesViewModel flows

MaelImhof:
# Summary

I'm currently working on using `HikesViewModel` everywhere in the app instead of `ListOfHikeRoutesViewModel` and `SavedHikesViewModel`, however I realize now that I need some more information made available by `HikesViewModel` to make the app work correctly with it.

To this end, I'm doing this intermediate PR to introduce the needed helper functions and new state flows.

# Details

Three parts of the code have been modified :

- `DeferredData.kt`: added helper functions
- `Hike.kt`: added helper functions
- `HikesViewModel.kt`: improved implementation and new public members

Tests have been added for `DeferredData` and `Hike`, and the existing ones for `HikesViewModel` were updated to reflect the changes.

## `DeferredData.kt`

- Added `DeferredData.obtained()` as a helper to determine whether an instance represents obtained data (it's shorter than doing `myVariable is DeferredData.Obtained` each time.
- Added `DeferredData.getOrNull()` to access the data or `null` if there is no data
- Added `DeferredData.getOrThrow()` to access the data or throw an exception if not available. This is to use when an `obtained()` check was done before.

## `Hike.kt`

- Copied `HikeRoute.getColor()` to `Hike.getColor()` for easier use
- Added `Hike.hasOsmData()` as a helper to check if a hike has a description, bounds and waypoints
- Added `Hike.withDetailsOrThrow()`, which asserts that all properties have been computed, and otherwise throws an exception. If everything was computed, returns an instance of `DetailedHike`, which is basically the same thing as the hike but without the `DeferredData` wrapper.

## `HikesViewModel.kt`

- Improved how `loading` is handled. Now, `loading` will be set to false only when there's no loading operation going on in the view model. This wasn't exactly the case before because of concurrency reasons.
- Added `allOsmDataLoaded`, a state flow that indicates whether all loaded hikes have their OSM data available. This is helpful to implement the Saved Hikes screen.
- Added `loadedHikesType`, a state flow that indicates whether saved hikes are loaded (saved hikes screen) or rather the current hikes are loaded from OSM (map screen)

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
@@ -48,4 +48,12 @@ sealed class DeferredData<out T> {
 
   /** Returns `true` if the data has been obtained, `false` otherwise. */
   fun obtained(): Boolean = this is Obtained
+
+  /** Returns the data if it has been obtained, otherwise returns `null`. */
+  fun getOrNull(): T? {
+    return when (this) {
+      is Obtained -> data
+      else -> null
+    }
+  }
 }
--- a/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
@@ -45,4 +45,7 @@ sealed class DeferredData<out T> {
    * job of [DeferredData].
    */
   data class Obtained<out T>(val data: T) : DeferredData<T>()
+
+  /** Returns `true` if the data has been obtained, `false` otherwise. */
+  fun obtained(): Boolean = this is Obtained
 }
--- a/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
@@ -56,4 +56,14 @@ sealed class DeferredData<out T> {
       else -> null
     }
   }
+
+  /**
+   * Returns the data if it has been obtained, otherwise throws an [IllegalStateException].
+   *
+   * @throws IllegalStateException If the data hasn't been obtained yet.
+   */
+  fun getOrThrow(): T {
+    check(this is Obtained)
+    return this.data
+  }
 }
--- a/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
@@ -68,24 +68,38 @@ data class Hike(
   }
 
   /**
+   * Indicates whether the hike has all of its data obtained.
+   *
+   * See [DeferredData] for more information about what it means for the data to be obtained.
+   *
+   * If this returns, true, you can safely call [withDetailsOrThrow] to get a [DetailedHike]
+   * instance
+   *
+   * @return True if all data were obtained for the hike, false otherwise.
+   */
+  fun isFullyLoaded(): Boolean =
+      description.obtained() &&
+          bounds.obtained() &&
+          waypoints.obtained() &&
+          elevation.obtained() &&
+          distance.obtained() &&
+          estimatedTime.obtained() &&
+          elevationGain.obtained() &&
+          difficulty.obtained()
+
+  /**
    * If all attributes of the hike are computed, casts everything to their respective data type and
    * returns a [DetailedHike] with the values to work with them directly without needing to perform
    * null checks.
+   *
+   * To make sure all details are loaded before calling this method, use [isFullyLoaded].
    *
    * If one or more attribute is missing, throws an [IllegalStateException].
    *
    * @throws IllegalStateException If one or more attribute of the hike has not been computed yet.
    */
   fun withDetailsOrThrow(): DetailedHike {
-    check(
-        description.obtained() &&
-            bounds.obtained() &&
-            waypoints.obtained() &&
-            elevation.obtained() &&
-            distance.obtained() &&
-            estimatedTime.obtained() &&
-            elevationGain.obtained() &&
-            difficulty.obtained())
+    check(isFullyLoaded())
     return DetailedHike(
         id,
         getColor(),
--- a/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
@@ -50,7 +50,12 @@ data class Hike(
   /** Helper to convert this [Hike] to a [SavedHike] object. */
   fun toSavedHike() = SavedHike(id, name ?: "", plannedDate)
 
-  /** Get the color of the route from its id. The color should be the same for the same route id. */
+  /**
+   * Get the color of the route from its id. The color should be the same for the same route id.
+   *
+   * @return The color as an integer. Use [androidx.compose.ui.graphics.Color] to convert it to an
+   *   actual color.
+   */
   fun getColor(): Int {
     return hikeColors[abs(id.hashCode()) % hikeColors.size]
   }
--- a/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
@@ -1,7 +1,9 @@
 package ch.hikemate.app.model.route
 
 import ch.hikemate.app.model.route.saved.SavedHike
+import ch.hikemate.app.ui.theme.hikeColors
 import com.google.firebase.Timestamp
+import kotlin.math.abs
 
 /**
  * Represents a hike route with associated information.
@@ -47,4 +49,9 @@ data class Hike(
 ) {
   /** Helper to convert this [Hike] to a [SavedHike] object. */
   fun toSavedHike() = SavedHike(id, name ?: "", plannedDate)
+
+  /** Get the color of the route from its id. The color should be the same for the same route id. */
+  fun getColor(): Int {
+    return hikeColors[abs(id.hashCode()) % hikeColors.size]
+  }
 }
--- a/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/Hike.kt
@@ -66,4 +66,62 @@ data class Hike(
   fun hasOsmData(): Boolean {
     return description.obtained() && bounds.obtained() && waypoints.obtained()
   }
+
+  /**
+   * If all attributes of the hike are computed, casts everything to their respective data type and
+   * returns a [DetailedHike] with the values to work with them directly without needing to perform
+   * null checks.
+   *
+   * If one or more attribute is missing, throws an [IllegalStateException].
+   *
+   * @throws IllegalStateException If one or more attribute of the hike has not been computed yet.
+   */
+  fun withDetailsOrThrow(): DetailedHike {
+    check(
+        description.obtained() &&
+            bounds.obtained() &&
+            waypoints.obtained() &&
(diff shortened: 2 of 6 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -155,27 +187,26 @@ class HikesViewModel(
...
    * @param onFailure Will be called if an error is encountered.
    */
   fun refreshSavedHikesCache(onSuccess: () -> Unit = {}, onFailure: () -> Unit = {}) =
       viewModelScope.launch {
         // Let the user know a heavy load operation is being performed
-        _loading.value = true
+        setLoading(true)
 
-        val success = refreshSavedHikesCacheAsync()
+        val success = refreshSavedHikesCacheAsync(forceOverwriteHikesList = false)
```
mchac15, line 201: Question: Could you explain why in this case `forceOverwriteHikeList` is set to false?

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -296,15 +327,7 @@ class HikesViewModel(
...
-      viewModelScope.launch {
-        // Notify the UI that a heavy loading operation is in progress
-        _loading.value = true
-
-        retrieveLoadedHikesOsmDataAsync(onSuccess, onFailure)
-
-        // The heavy loading operation is done now
-        _loading.value = false
-      }
+      viewModelScope.launch { retrieveLoadedHikesOsmDataAsync(onSuccess, onFailure) }
```
mchac15, line 330: Question: could you explain why `setLoading()` is not called here?

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -391,9 +451,39 @@ class HikesViewModel(
...
+   * Internal helper function.
+   *
+   * Sets [_loadedHikesType] with the new provided value if not already equal to that value.
+   *
+   * Requires [_hikesMutex] to have been acquired by the caller.
+   */
+  private fun setLoadedHikesType(value: LoadedHikes) {
+    if (_loadedHikesType.value != value) {
+      _loadedHikesType.value = value
+    }
```
mchac15, line 486: Question: I am curious as to why you check if `_loadedHikesType` isn't equal to the `value` given as parameter?

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -478,7 +568,7 @@ class HikesViewModel(
    *
    * @return True if the operation is successful, false otherwise.
```
mchac15, line 569: Nitpick:
```suggestion
   * @param forceOverwriteHikesList
   *
   * @return True if the operation is successful, false otherwise.
```

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -478,7 +568,7 @@ class HikesViewModel(
...
    */
-  private suspend fun refreshSavedHikesCacheAsync(): Boolean =
+  private suspend fun refreshSavedHikesCacheAsync(forceOverwriteHikesList: Boolean): Boolean =
```
mchac15, line 571: Suggestion: Maybe here you could add why exactly the `forceOverwriteHikesList` parameter is used for, since it isn't really clear.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -70,25 +70,57 @@ class HikesViewModel(
...
   private val _selectedHike = MutableStateFlow<Hike?>(null)
 
-  /** Internal type used to store where the hikes in [_hikeFlowsMap] come from. */
-  private enum class LoadedHikes {
+  /**
+   * Enum used to designate where the hikes loaded in [hikeFlows] come from semantically.
+   *
+   * See the possible values for this enum along with [loadedHikesType] for more information.
+   */
+  enum class LoadedHikes {
```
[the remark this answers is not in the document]
mchac15: Answer: I am guessing the reason is testing, after looking at the tests.

mchac15 (changes requested):
After reading the code again and again, I have to say your code is very well commented and written. Most of my comments are nitpicks and potential improvements, that sometimes depend on the intended use of your code when it's used in the UI.
Other than that thanks again for having such well documented code

app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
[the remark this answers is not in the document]
MaelImhof: This is a great idea, I thought of something similar though I think you went further in your reasoning. There is a task for this called [Handle errors in new HikesViewModel](https://github.com/orgs/HikeMate/projects/1/views/5?pane=issue&itemId=88094162) and I think this is hence out of scope for this task, however we should indeed consider it, and I'll add it to the task's description.

app/src/main/java/ch/hikemate/app/model/route/Hike.kt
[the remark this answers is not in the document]
MaelImhof: I did not add it because I only used it in `withDetailsOrThrow`, but it might be a good idea indeed. I added it in a local commit, will push when I finish reading the review 👌🏻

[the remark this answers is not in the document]
MaelImhof: I don't think this is necessary. `Hike` represents a hike that is not necessarily fully loaded. `DetailedHike` is for sure fully loaded, so you never want to go back to a `Hike`.

app/src/main/java/ch/hikemate/app/model/route/Hike.kt
```diff
@@ -47,4 +49,79 @@ data class Hike(
 ) {
   /** Helper to convert this [Hike] to a [SavedHike] object. */
   fun toSavedHike() = SavedHike(id, name ?: "", plannedDate)
+
+  /** Get the color of the route from its id. The color should be the same for the same route id. */
```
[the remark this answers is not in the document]
MaelImhof: Done 👌🏻

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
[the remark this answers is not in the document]
MaelImhof: Actually it's mostly because of `loadedHikesType`. In my previous implementation of `HikesViewModel`, this enum was used for an internal state, and it did not need to be known outside.

However I realized this state would be useful outside of `HikesViewModel`, so I created the state flow `loadedHikesType`.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: This is related to the `LoadedHikes` enum. This enum is used to maintain state about what type of hikes are currently loaded in the hikes list.

- If we are on the saved hikes screen, that state will be `LoadedHikes.FromSaved`
- If we are in the map screen and we searched for hikes, it will be `LoadedHikes.FromBounds`

Now, even when we are on the map screen, we want to load the saved hikes in the background, to know which ones of the OSM hikes are saved and which ones are not.

To do this, the view model keeps a cache of the user's saved hikes at all times. `refreshSavedHikesCache` is used to load the saved hikes again, but not necessarily overwrite the current list of hikes.

For example, we searched for hikes on the map and we want to see which ones are saved, but we are not sure our local cache of saved hikes is up-to-date. We can call `refreshSavedHikesCache`, and it will refresh the cache without unloading the searched hikes.

Now, sometimes you want to actually overwrite the list of loaded hikes with the saved hikes (when you go on the Saved Hikes screen). In those times you call `loadSavedHikes`, which sets `forceOverwriteHikesList` to `true`.

Here, it's just a call to `refreshSavedHikesCache`, so we don't want to overwrite the existing hikes if any. So we set it to `false`.

Sorry for the lengthy explanation, hope that clears things out!

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: Good point, I moved it to inside of `retrieveLoadedHikesOsmDataAsync` directly. This is because I wanted to set `loading` to `true` if and only if a loading operation was actually performed. The method first does some checks about whether the loading operation is necessary, and performs an early return if the request is not needed.

So I only call `setLoading` when the request is actually being performed.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: I'm not sure about this, but from what I've seen, setting a state flow to the same value as its previous will still launch UI updates, even though unnecessary.

So to avoid unnecessary updates I check the current value first.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: Done 👌🏻

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
MaelImhof: I added some docstring to help 👌🏻

app/src/main/java/ch/hikemate/app/model/route/DeferredData.kt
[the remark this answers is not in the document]
mchac15: I agree

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
mchac15: Thanks for adding the documentation for this since it wasn't really clear. Thanks for the explanation the implementation looks good to me.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
mchac15: I understand why you did it, but it seems to me that that is inconsistent with the rest of the code since I see that you always call the `setLoading` function in the wrapper functions and not in the `async` functions.
But if there isn't any way to solve this I think you can leave it as is, since the code is robust enough anyways since it handles the case where there is an error well: taking into account the `setLoading(false)` call is outside the `withLock` call, so anyways the code would be equivalent.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
mchac15: Yes, you are right I looked it up and even if it's the same value it will trigger a re composition.

mchac15 (changes requested):
Just one more conversation for the sake of consistency of the code and you are good to go. 
I want to thank you especially for taking the time to make lengthy and thoughtful answers to my comments since they help me understand your code better.

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
mchac15: I think we can leave it like this, it doesn't pose any real problems.

mchac15 (approved):
Thanks for your patience by answering to my questions to understand and review your code. You can merge now.

PR #235, opened 2024-12-01
feat: add mapstate that saves state of the map screen map

davidholzwarth:
# Summary

Added a `MapState` to the list of hikes vM, that saves the center and zoom level of the map whenever the user navigates back from hike details screen -> map screen. This state is then used the next time the mapscreen is loaded, which means that whenever you navigate from HD to MS, the state of the map is preserved.

# Details

The MapView is declared so early, since it needs to be declared before the LaunchedEffect is declared. The initialitsation still happens where it used to, just using a MutableState, so that the LaunchedEffect has knowledge of the mapView just before a back-navigation happen. 

There are some adjustments to the hikeVM and mapscreen too to accomodate the new mapstate, notably instead of directly initialising the map's initial value in the MapScreen, they are now initialized directly in the hVM.


app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -404,6 +408,31 @@ class HikesViewModel(
...
+   * Fetches the center and zoom level of the map's. This function enables saving the map's state,
+   * so that the map's center and zoom level can be preserved when navigating between screens.
+   *
+   * @return The current [MapUtils.MapViewState] containing the map's center point and zoom level.
+   */
+  fun getMapState(): MapUtils.MapViewState {
+    return _mapState.value
+  }
+
   /**
```
rBergCode, line 436: Praise: That's clean. Well done!

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -264,4 +265,9 @@ object MapUtils {
         mapView.zoomLevelDouble,
         MapScreen.CENTER_MAP_ANIMATION_TIME)
   }
+
+  data class MapViewState(
+      val center: GeoPoint? = MapInitialValues().mapInitialCenter,
```
rBergCode, line 270: Question: Is there a reason to use `GeoPoint` instead of `LatLng` here ?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -132,9 +134,22 @@ fun HikeDetailScreen(
 
   val selectedHike by hikesViewModel.selectedHike.collectAsState()
 
+  // Gets initialized here so that the LaunchedEffect has access to it. The value is only actually
+  // initialised and used in HikeDetailsContent
+  val mapViewState = remember { mutableStateOf<MapView?>(null) }
```
rBergCode, line 139: Suggestion: Would that work to remove the optional and initialize it to `hikeDetailsMap` right away ? In the `LaunchedEffect` in order to have access to `selectedHike`.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -132,9 +134,22 @@ fun HikeDetailScreen(
...
+
+  // If the selected hike is null, save the map's state and go back to the map screen
```
rBergCode, line 141: Suggestion: You can add a Launched effect directly in `HikeDetailsContent`

rBergCode (changes requested):
Thanks for the fix. The app looks more polished now.
Just to make sure we are on the same page: Was the expected behavior to recenter the user to its last location on map screen or on the location of the hike ?

Also, I have few suggestions in `HikeDetailsScreen`.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
davidholzwarth: Hmmm. How would that work? If we declare the variable inside of the Launched Effect, then its not in the global scope, so we would need to do some kind of lateinit, or? Also, we need the selectedHike to initialize the mapView, and that may still be null at this point in the code. The hike is only null-checked in the LaunchedEffect, however we need the mapState to implement the LaunchedEffect. This is a very awkward circualr dependency, but if you can think of a cleaner way to fix it I'd definitely be open to, I don't particularly love this solution either....

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
davidholzwarth: I am not sure that that will work. We could pass `selectedHike` as a parameter to the `HikeDetailsContent`, and have the `LaunchedEffect` there, but then all the code that gets executed before the `LaunchedEffect`, especially this line: `val detailedHike = selectedHike!!.withDetailsOrThrow()` will execute when selectedHike may be null.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
MaelImhof: @rBergCode asked me to help on this particular topic since I refactored the code in the first place.

The problem here is that :

- `LaunchedEffect(selectedHike)` is in the top composable function (top-level)
- When the `selectedHike` changes, we should save the map's state before leaving the screen
- To save the map's state, you need access to the instance of the map
- The map is defined at a lower level

So what you did is you brought the `mapView` instance up in the composable hierarchy. While this works, we might be able to avoid making `mapView` nullable.

To do this, instead of editing the existing `LaunchedEffect`, perhaps you could create a new `LaunchedEffect(selectedHike)` in `HikeDetailsContent` directly, where the map instance is created. Once it is created, you can execute the LaunchedEffect without any issue, and no need for null checks.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
MaelImhof: See my comment in the above conversation.

davidholzwarth:
@rBergCode its to save the last state of the user, so if he is on the map or not doesnt matter, its just to avoid the map moving around a lot when the user navigates back from the HD screen

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
davidholzwarth: Do you mean to create a second LaunchedEffect or move the existing one down?

MaelImhof: Create a second one.

davidholzwarth: But then the first one in the higher level will execute first, im pretty sure, and once the back navigation is triggered the second one will never get a change to execute.

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
davidholzwarth: I did it since `MapInitialValues` uses GeoPoint

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
MaelImhof: Could you try it out maybe? I don't think it will cause any particular issues, at least I hope not.

davidholzwarth: I have and it did not work

MaelImhof: I tried it out on my side too and it indeed does not work, so let's keep this way of doing. Sorry for bothering you!

rBergCode (dismissed):
From the discussion, you can merge. Thanks for the job!

rBergCode (changes requested):
Small fix about the merge and you can finish this!

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -180,13 +196,16 @@ fun HikeDetailScreen(
 @Composable
 fun HikeDetailsContent(
     hike: DetailedHike,
+    mapViewState: MutableState<MapView?>,
     navigationActions: NavigationActions,
     hikesViewModel: HikesViewModel,
     userHikingLevel: HikingLevel
 ) {
+  BackHandler { hikesViewModel.unselectHike() }
+
```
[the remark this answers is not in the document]
davidholzwarth: removed

davidholzwarth (commented):
i've removed the guilty line

rBergCode (approved):
Thanks! Good for merging now!

davidholzwarth:
@rBergCode sorry for re-requesting, but I've added some more test. Could you have a look at them please?

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
```diff
@@ -351,4 +353,15 @@ class MapScreenTest : TestCase() {
...
+  @Test
+  fun mapLoadsFromMapState() {
+    // Uses Mockk since Mockito can not mock/spy final classes
+    val spyHikesViewModel = spyk(hikesViewModel)
+    setUpMap(hikesViewModel = spyHikesViewModel)
+
+    composeTestRule.waitForIdle()
+
+    io.mockk.verify { spyHikesViewModel.getMapState() }
+  }
```
rBergCode, line 366: Question: Could you explain what does `spyk` do exactly? What's the difference with `mock` ?

Also, Why have you added this test ?

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -397,4 +398,19 @@ class HikeDetailScreenTest {
...
+
+    composeTestRule.onNodeWithTag(BACK_BUTTON_TEST_TAG).performClick()
+    composeTestRule.waitForIdle()
+
+    val zoomLevel = MapUtils.calculateBestZoomLevel(detailedHike.bounds).toDouble()
+    val center = MapUtils.getGeographicalCenter(detailedHike.bounds)
+    assertEquals(zoomLevel, hikesViewModel.getMapState().zoom, 0.1)
+    assertEquals(center.latitude, hikesViewModel.getMapState().center.latitude, 0.1)
+    assertEquals(center.longitude, hikesViewModel.getMapState().center.longitude, 0.1)
+  }
```
rBergCode, line 415: Praise: Clean!

rBergCode (commented):
I just have a question. Thanks for the new tests!

app/src/androidTest/java/ch/hikemate/app/ui/map/MapScreenTest.kt
davidholzwarth: The reason I spy here is because if I don't I can not call `verify` on the `hikesViewModel`, since no one keeps track of how often it was called. By using spy, I can keep track of how often the `.getMapState()` was called, but I don't need to go through the effort of mocking the `HikesViewModel`, since it just uses its original functionality.

davidholzwarth: I added the test to ensure that the MapScreen actually calls and makes use of the newly added `MapState`. I noticed that I had not tested this before. Imo its a useful test to ensure that the functionality I implemented works continuously.

rBergCode: Thanks for the explanation. More clearer to me now.

rBergCode (dismissed):
Thanks for the explanation!

rBergCode (approved):
Still good after 3 merge with main :)

PR #237, opened 2024-12-02
feat: add details to run hike screen

davidholzwarth:
# Summary

Added the hike details to the run hike screen bottom sheet as per the Figma design. This includes fetching the elevation data, however this may have to be undone not that the hike classes are unified. Mael agreed to look at this screen once it it merged. Currently, the "current elevation" and "distance progressed" information is hard-coded, since we do not have a location viewModel that provides the UI with location updates. This is being worked on my Rémy.

# Notes
1. I made the naming structure of the run hike screen test tags more clear, since as adding tests for the new features I noticed in this PR that they were quite convoluted. 
2. When adding the stop this hike button I also improved the back-navigation to be more crash resistant, by making the stop this run button and the back button update a `wantToNavigateBack` boolean to false, and on this change to false the back navigation happens. This variable can only be set to false, so even when spamming the buttons, only one back navigation is triggered.

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
@@ -149,6 +149,7 @@ class RunHikeScreenTest {
     composeTestRule.onNodeWithTag(TEST_TAG_BOTTOM_SHEET).assertExists()
   }
 
+  /** Test all details which do not change during the hike run. */
   @Test
   fun runHikeScreen_displaysStaticDetails() = runTest {
     composeTestRule.setContent {
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
@@ -1,7 +1,11 @@
 package ch.hikemate.app.ui.map
 
+import androidx.compose.ui.test.assert
+import androidx.compose.ui.test.assertCountEquals
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onAllNodesWithTag
 import androidx.compose.ui.test.onNodeWithTag
 import ch.hikemate.app.model.elevation.ElevationService
 import ch.hikemate.app.model.profile.HikingLevel
@@ -9,13 +13,19 @@ import ch.hikemate.app.model.profile.Profile
 import ch.hikemate.app.model.profile.ProfileRepository
 import ch.hikemate.app.model.profile.ProfileViewModel
 import ch.hikemate.app.model.route.Bounds
+import ch.hikemate.app.model.route.DetailedHikeRoute
+import ch.hikemate.app.model.route.HikeDifficulty
 import ch.hikemate.app.model.route.HikeRoute
 import ch.hikemate.app.model.route.HikeRoutesRepository
 import ch.hikemate.app.model.route.LatLong
 import ch.hikemate.app.model.route.ListOfHikeRoutesViewModel
+import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_DETAIL_ROW_TAG
 import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_BACK_BUTTON
 import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_BOTTOM_SHEET
+import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_ELEVATION_GRAPH
+import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_HIKE_NAME
 import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_MAP
+import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_TOTAL_DISTANCE_TEXT
 import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_ZOOM_BUTTONS
 import ch.hikemate.app.ui.navigation.NavigationActions
 import com.google.firebase.Timestamp
@@ -48,6 +58,14 @@ class RunHikeScreenTest {
           name = "Sample Hike",
           description =
               "A scenic trail with breathtaking views of the Matterhorn and surrounding glaciers.")
+
+  private val detailedRoute =
+      DetailedHikeRoute(
+          route = route,
+          totalDistance = 13.543077559212616,
+          elevationGain = 68.0,
+          estimatedTime = 169.3169307105514,
+          difficulty = HikeDifficulty.DIFFICULT)
 
   private val profile =
       Profile(
@@ -115,7 +133,7 @@ class RunHikeScreenTest {
   }
 
   @Test
-  fun runHikeScreen_displaysBottomSheet() {
+  fun runHikeScreen_displaysBottomSheet() = runTest {
     composeTestRule.setContent {
       RunHikeScreen(
           listOfHikeRoutesViewModel = listOfHikeRoutesViewModel,
@@ -125,4 +143,40 @@ class RunHikeScreenTest {
     }
     composeTestRule.onNodeWithTag(TEST_TAG_BOTTOM_SHEET).assertExists()
   }
+
+  @Test
+  fun runHikeScreen_displaysStaticDetails() = runTest {
+    composeTestRule.setContent {
+      RunHikeScreen(
+          listOfHikeRoutesViewModel = listOfHikeRoutesViewModel,
+          profileViewModel = profileViewModel,
+          navigationActions = mockNavigationActions,
+      )
+    }
+
+    composeTestRule.onNodeWithTag(TEST_TAG_HIKE_NAME).assertIsDisplayed()
+
+    composeTestRule.onAllNodesWithTag(TEST_TAG_DETAIL_ROW_TAG).assertCountEquals(4)
+    composeTestRule.onAllNodesWithTag(TEST_TAG_DETAIL_ROW_TAG).assertCountEquals(4)
+
+    composeTestRule
+        .onNodeWithTag(TEST_TAG_TOTAL_DISTANCE_TEXT)
+        .assertIsDisplayed()
+        .assert(hasText("13.54km"))
+  }
+
+  @Test
+  fun runHikeScreen_displaysElevationGraph() = runTest {
+    composeTestRule.setContent {
+      RunHikeScreen(
+          listOfHikeRoutesViewModel = listOfHikeRoutesViewModel,
+          profileViewModel = profileViewModel,
+          navigationActions = mockNavigationActions,
+      )
+    }
+
+    composeTestRule.onNodeWithTag(TEST_TAG_ELEVATION_GRAPH).assertIsDisplayed()
+  }
+
+  @Test fun runHikeScreen_displaysStopHikeButton() = runTest {}
 }
--- a/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
@@ -18,16 +18,6 @@ import ch.hikemate.app.model.route.HikeRoute
 import ch.hikemate.app.model.route.HikeRoutesRepository
 import ch.hikemate.app.model.route.LatLong
 import ch.hikemate.app.model.route.ListOfHikeRoutesViewModel
-import ch.hikemate.app.ui.map.HikeDetailScreen.TEST_TAG_DETAIL_ROW_TAG
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_BACK_BUTTON
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_BOTTOM_SHEET
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_ELEVATION_GRAPH
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_HIKE_NAME
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_MAP
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_PROGRESS_TEXT
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_STOP_HIKE_BUTTON
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_TOTAL_DISTANCE_TEXT
-import ch.hikemate.app.ui.map.RunHikeScreen.TEST_TAG_ZOOM_BUTTONS
 import ch.hikemate.app.ui.navigation.NavigationActions
 import com.google.firebase.Timestamp
 import kotlinx.coroutines.ExperimentalCoroutinesApi
@@ -100,7 +90,7 @@ class RunHikeScreenTest {
           navigationActions = mockNavigationActions,
       )
     }
-    composeTestRule.onNodeWithTag(TEST_TAG_MAP).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(RunHikeScreen.TEST_TAG_MAP).assertIsDisplayed()
   }
 
   @Test
@@ -112,7 +102,7 @@ class RunHikeScreenTest {
           navigationActions = mockNavigationActions,
       )
     }
-    composeTestRule.onNodeWithTag(TEST_TAG_BACK_BUTTON).assertIsDisplayed()
+    composeTestRule.onNodeWithTag(RunHikeScreen.TEST_TAG_BACK_BUTTON).assertIsDisplayed()
   }
 
   @Test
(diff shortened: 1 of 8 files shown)


app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+            val elevationGainString = hikeRoute.elevationGain.roundToInt().toString()
+            val hourString =
+                String.format(Locale.getDefault(), "%02d", (hikeRoute.estimatedTime / 60).toInt())
+            val minuteString =
+                String.format(
+                    Locale.getDefault(), "%02d", (hikeRoute.estimatedTime % 60).roundToInt())
+
+            DetailRow(
+                label = stringResource(R.string.run_hike_screen_label_current_elevation),
+                value = "50m")
```
MaelImhof, line 266: Suggestion: move this string to `strings.xml` as `"%dm"` and use `50` as a value here instead.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+            DetailRow(
+                label = stringResource(R.string.run_hike_screen_label_elevation_gain),
+                value = "${elevationGainString}m")
```
MaelImhof, line 269: Suggestion: move this string to `strings.xml` as `"%dm"`.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+            DetailRow(
+                label = stringResource(R.string.run_hike_screen_label_estimated_time),
+                value =
+                    if (hikeRoute.estimatedTime / 60 < 1) "${minuteString}min"
+                    else "${hourString}h${minuteString}")
```
MaelImhof, line 274: Suggestion: move those to `strings.xml` as `"%dmin"` and `"%dh%d"`.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+                    Color(
+                        ContextCompat.getColor(
+                            LocalContext.current, hikeRoute.difficulty.colorResourceId)),
+            )
+
+            BigButton(
+                buttonType = ButtonType.PRIMARY,
+                label = stringResource(R.string.run_hike_screen_stop_run_button_label),
+                onClick = onStopTheRun,
+                modifier = Modifier.padding(top = 16.dp).testTag(TEST_TAG_STOP_HIKE_BUTTON),
```
MaelImhof, line 288: Nitpick: can you use `RunHikeScreen.` rather than importing the tag itself?

```suggestion
                modifier = Modifier.padding(top = 16.dp).testTag(RunHikeScreen.TEST_TAG_STOP_HIKE_BUTTON),
```

MaelImhof (changes requested):
Nice work ! There are a lot of comments, but they are mostly "move this to `strings.xml`" or "use complete name for test tags". Your tests are good too 👌🏻

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
```diff
@@ -111,18 +124,104 @@ class RunHikeScreenTest {
...
+          listOfHikeRoutesViewModel = listOfHikeRoutesViewModel,
+          profileViewModel = profileViewModel,
+          navigationActions = mockNavigationActions,
+      )
+    }
+
+    composeTestRule.onNodeWithTag(TEST_TAG_HIKE_NAME).assertIsDisplayed()
+
+    composeTestRule.onAllNodesWithTag(TEST_TAG_DETAIL_ROW_TAG).assertCountEquals(4)
+    composeTestRule.onAllNodesWithTag(TEST_TAG_DETAIL_ROW_TAG).assertCountEquals(4)
```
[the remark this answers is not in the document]
davidholzwarth: Thats actually a good catch, one of them is supposed to be detail_row_value, instead of tag. Thanks!

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+                horizontalArrangement = Arrangement.SpaceBetween) {
+                  Text(
+                      text = stringResource(R.string.run_hike_screen_zero_distance_progress_value),
+                      style = MaterialTheme.typography.bodyLarge,
+                      fontWeight = FontWeight.Bold,
+                      textAlign = TextAlign.Left,
+                      modifier = Modifier.padding(top = 8.dp),
+                  )
+                  Text(
+                      text = "23% complete",
```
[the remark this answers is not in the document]
davidholzwarth: Definitely, good idea

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+                      style = MaterialTheme.typography.bodyLarge,
+                      color = hikeColor,
```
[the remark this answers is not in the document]
davidholzwarth: Ahh, yes I thought about that too, but I thought to myself, that it would be more aesthetically unified and pleasing to have "everything" in the `hikeColor`. Don't you think the user will realise that this does not indicate an error? Because perhaps we could also consider re-thinking our color choices to not include red. If it really bothers you I can definitively hard-code it to green or some other color, but I think this solution with the `hikeColor`s is quite satisfying, since these colors also all work in dark & light mode.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+                  Text(
+                      text =
+                          stringResource(R.string.run_hike_screen_distance_progress_value_format)
+                              .format(hikeRoute.totalDistance),
+                      style = MaterialTheme.typography.bodyLarge,
+                      fontWeight = FontWeight.Bold,
+                      textAlign = TextAlign.Right,
+                      modifier = Modifier.padding(top = 8.dp).testTag(TEST_TAG_TOTAL_DISTANCE_TEXT),
+                  )
+                }
```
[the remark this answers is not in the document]
davidholzwarth: Yes! Smart 💯

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+                value =
+                    if (hikeRoute.estimatedTime / 60 < 1) "${minuteString}min"
+                    else "${hourString}h${minuteString}")
+            DetailRow(
+                label = stringResource(R.string.run_hike_screen_label_difficulty),
+                value = stringResource(hikeRoute.difficulty.nameResourceId),
+                valueColor =
+                    Color(
+                        ContextCompat.getColor(
+                            LocalContext.current, hikeRoute.difficulty.colorResourceId)),
```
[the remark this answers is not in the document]
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -156,6 +199,95 @@ fun RunHikeBottomSheet(
...
+            )
+
+            BigButton(
+                buttonType = ButtonType.PRIMARY,
+                label = stringResource(R.string.run_hike_screen_stop_run_button_label),
+                onClick = onStopTheRun,
+                modifier = Modifier.padding(top = 16.dp).testTag(TEST_TAG_STOP_HIKE_BUTTON),
+                fillColor = Color(0xFFE83B3D))
```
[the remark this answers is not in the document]
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -263,30 +256,41 @@ fun RunHikeBottomSheet(
 
             DetailRow(
                 label = stringResource(R.string.run_hike_screen_label_current_elevation),
-                value = "50m")
+                // TODO hardcoded to 50m for now
+                value =
+                    stringResource(R.string.run_hike_screen_value_format_current_elevation)
+                        .format(50))
```
MaelImhof, line 262: Nitpick: you actually don't need to call `format` manually, you can just pass the arguments directly to `stringResource`, it calls `format` under the hood.

```suggestion
                    stringResource(R.string.run_hike_screen_value_format_current_elevation, 50))
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -263,30 +256,41 @@ fun RunHikeBottomSheet(
...
             DetailRow(
                 label = stringResource(R.string.run_hike_screen_label_elevation_gain),
-                value = "${elevationGainString}m")
+                value =
+                    stringResource(R.string.run_hike_screen_value_format_elevation_gain)
+                        .format(hikeRoute.elevationGain.roundToInt()))
```
MaelImhof, line 267: Nitpick: you actually don't need to call `format` manually, you can just pass the arguments directly to `stringResource`, it calls `format` under the hood.

```suggestion
                    stringResource(R.string.run_hike_screen_value_format_elevation_gain, hikeRoute.elevationGain.roundToInt()))
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -250,11 +244,10 @@ fun RunHikeBottomSheet(
...
                       textAlign = TextAlign.Right,
-                      modifier = Modifier.padding(top = 8.dp).testTag(TEST_TAG_TOTAL_DISTANCE_TEXT),
+                      modifier = Modifier.testTag(RunHikeScreen.TEST_TAG_TOTAL_DISTANCE_TEXT),
                   )
                 }
 
-            val elevationGainString = hikeRoute.elevationGain.roundToInt().toString()
             val hourString =
                 String.format(Locale.getDefault(), "%02d", (hikeRoute.estimatedTime / 60).toInt())
             val minuteString =
```
MaelImhof, line 253: Change: these variables are now unused, my suggestion would be to create `val hours` and `val minutes` and use them below instead of re-computing every time.

```suggestion
            val hours = hikeRoute.estimatedTime / 60
            val minutes = hikeRoute.estimatedTime % 60
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -263,30 +256,41 @@ fun RunHikeBottomSheet(
...
+                    stringResource(R.string.run_hike_screen_value_format_elevation_gain)
+                        .format(hikeRoute.elevationGain.roundToInt()))
             DetailRow(
                 label = stringResource(R.string.run_hike_screen_label_estimated_time),
                 value =
-                    if (hikeRoute.estimatedTime / 60 < 1) "${minuteString}min"
-                    else "${hourString}h${minuteString}")
+                    if (hikeRoute.estimatedTime / 60 < 1)
+                        stringResource(R.string.run_hike_screen_value_format_estimated_time_minutes)
+                            .format((hikeRoute.estimatedTime % 60).roundToInt())
```
MaelImhof, line 273: Nitpick: you actually don't need to call `format` manually, you can just pass the arguments directly to `stringResource`, it calls `format` under the hood.

```suggestion
                        stringResource(R.string.run_hike_screen_value_format_estimated_time_minutes, (hikeRoute.estimatedTime % 60).roundToInt())
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -263,30 +256,41 @@ fun RunHikeBottomSheet(
...
+                    else
+                        stringResource(
+                                R.string
+                                    .run_hike_screen_value_format_estimated_time_hours_and_minutes)
+                            .format(
+                                (hikeRoute.estimatedTime / 60).toInt(),
+                                (hikeRoute.estimatedTime % 60).roundToInt()))
```
MaelImhof, line 280: Nitpick: you actually don't need to call `format` manually, you can just pass the arguments directly to `stringResource`, it calls `format` under the hood.

```suggestion
                        stringResource(
                                R.string
                                    .run_hike_screen_value_format_estimated_time_hours_and_minutes, (hikeRoute.estimatedTime / 60).toInt(),
                                (hikeRoute.estimatedTime % 60).roundToInt()))
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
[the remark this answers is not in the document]
MaelImhof: True, I personally like it as well. Maybe we can leave it as-is for now and discuss it for a future PR if we want to change it, but it should be discussed with the team and the stakeholders.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -263,30 +256,41 @@ fun RunHikeBottomSheet(
 
             DetailRow(
```
MaelImhof, line 257: Suggestion: I noticed only now that you use the `DetailRow` component of the `HikeDetailsScreen`, should it become a component of its own in the `components` package (creating a new file `DetailRow.kt` with the composable inside, moving the test tags to that file, and so on)?

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
```diff
@@ -200,6 +200,10 @@ fun HikeDetailScreen(
     profileViewModel.getProfileById(authViewModel.currentUser.value!!.uid)
   }
 
+  var wantToRunHike = false
+
+  LaunchedEffect(wantToRunHike) { navigationActions.navigateTo(Screen.RUN_HIKE) }
+
```
MaelImhof, line 206: Change: I'm not sure why you added this, but now the details screen always switches immediately to the run hike screen, because the `LaunchedEffect` will be executed once on the first composition. What was the goal?

MaelImhof (changes requested):
Thank you for the nitpick changes, a few comments left but mostly, an addition you made actually broke the details screen. Not sure what the exact goal of this addition was, see discussion.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailScreen.kt
davidholzwarth: Yup, it should be fixed now, I made the launch effect check whether `wantToRunHike` is actually positive before it starts the navigation. Without this feature, spamming the run this hike button triggers multiple navigations to the run hike screen, which causes it to get launched multiple times...

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: fair point, I've moved it and updated the test tags

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
davidholzwarth: done

MaelImhof:
We currently have a situation with the elevation API we are using, their servers systematically fail and timeout on every request. It might be temporary, but it does not seem to come from our code.

Hence, the CI might not be passing because of that.

MaelImhof (approved):
Very nice, looks good to me ! Let's see if any test fails that shouldn't fail, but if the failure is related to the elevation API, we can ignore it.

Issue #224, opened 2024-11-29
Elevation graphs and difference altitude don't seem to work


Wylwi:
As commented in the MS2 feedback:
_Elevation graphs don’t seem to work; their status changes from ‘loading graph’ to ‘no data.’_
This most probably an issue in the repository or even from the API.

PR #246, opened 2024-12-05
chore: sanitize inputs to limit their length

rBergCode:
# Summary
The feedback from M2 mentioned:

> All user inputs should be sanitized to prevent long strings from breaking the UI. Currently, someone can input a very long name while signing up.

As a result, I limit the number of character of all inputs in the app.

# Details
To prevent reuse of code, I refactored an input in a new `CustomTextField` in order to have this configuration in one place only.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
@@ -1,0 +1,75 @@
+package ch.hikemate.app.ui.components
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.text.selection.TextSelectionColors
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.OutlinedTextFieldDefaults
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
+import androidx.compose.ui.text.style.TextAlign
+import ch.hikemate.app.ui.theme.primaryColor
+
+object CustomTextField {
+  const val MAX_NAME_LENGTH = 40
+  const val MAX_EMAIL_LENGTH = 50
+  const val MAX_PASSWORD_LENGTH = 50
+}
+
+/**
+ * Custom text field with custom colors and optional max length.
+ *
+ * @param value The value of the text field.
+ * @param onValueChange The callback when the value changes.
+ * @param label The label of the text field.
+ * @param isPassword Whether the text field is a password field.
+ * @param maxLength The maximum length of the text field.
+ * @param modifier The modifier of the text field.
+ */
+@Composable
+fun CustomTextField(
+    value: String,
+    onValueChange: (String) -> Unit,
+    label: String,
+    isPassword: Boolean = false,
+    maxLength: Int? = null,
+    modifier: Modifier = Modifier,
+) {
+
+  OutlinedTextField(
+      modifier = modifier.fillMaxWidth(),
+      colors =
+          OutlinedTextFieldDefaults.colors()
+              .copy(
+                  focusedLabelColor = primaryColor,
+                  focusedIndicatorColor = primaryColor,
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Custom text field with custom colors and optional max length.
+ *
+ * @param value The value of the text field.
+ * @param onValueChange The callback when the value changes.
+ * @param label The label of the text field.
+ * @param modifier The modifier of the text field.
+ * @param isPassword Whether the text field is a password field.
+ * @param maxLength The maximum length of the text field.
+ */
```
Wylwi, line 30: Nitpick: Could you add a comment or an option for the `singleLine = true`? Like, either allowing the dev to change it, or specify that it's always a one line text field

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
```diff
@@ -0,0 +1,75 @@
...
+      supportingText =
+          if (maxLength != null) {
+            {
+              Text(
+                  text = "${value.length} / $maxLength",
+                  modifier = Modifier.fillMaxWidth(),
+                  textAlign = TextAlign.End,
+              )
+            }
+          } else null,
```
Wylwi, line 71: Praise: That's a nice addition!

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -120,23 +118,13 @@ fun EditProfileScreen(
...
-                                  handleColor = primaryColor,
-                                  backgroundColor = primaryColor,
-                              )),
+          CustomTextField(
               value = name,
               onValueChange = { name = it },
-              label = { Text(context.getString(R.string.profile_screen_name_label)) },
-              singleLine = true)
+              label = context.getString(R.string.profile_screen_name_label),
+              maxLength = 40,
```
Wylwi, line 125: Warning: You should use the `CustomTextField.MAX_NAME_LENGTH`

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
+import androidx.compose.ui.text.style.TextAlign
+import ch.hikemate.app.ui.theme.primaryColor
+
+object CustomTextField {
+  const val MAX_NAME_LENGTH = 40
+  const val MAX_EMAIL_LENGTH = 50
+  const val MAX_PASSWORD_LENGTH = 50
+}
```
Wylwi, line 19: Question: How have you defined those constants?

Wylwi (changes requested):
That's a nice addition, and will definitely improve the app stability! I just have a few questions and nitpicks!

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
rBergCode: Completly arbitrary. Do you have a better value in mind ?

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
rBergCode: Yep okay

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
Wylwi: Nop, it was just to know!

app/src/main/java/ch/hikemate/app/ui/components/CustomTextField.kt
Wylwi: Thanks!

Wylwi (approved):
Thanks for the changes! You're good to go for merging now!

Wylwi:
(CI are failing because of elevation issues, which are going to be fixed in #224)

PR #248, opened 2024-12-05
feat: add facilities to the RunHike and HikeDetails screens

mchac15:
# Summary
This pull request includes significant updates primarily focused on the addition of facilities-related functionality in the `HikeDetailsScreen` and `RunHikeScreen`. The code dynamically filters and renders facilities based on map zoom levels and proximity to hiking paths.

## Implementation Details
The system consists of three main components working together:

### `FacilitiesViewModel`
Manages the core filtering logic determining which facilities to show based on:
- Map zoom level thresholds
- Distance from the active route 
- Maximum facility count per zoom level
- Dynamic updates on map interaction

### Map Interaction System
- Uses debounced `StateFlows` to handle scroll/zoom events efficiently
- Carefully manages `MapView` lifecycle
- Prevents excessive updates while maintaining responsiveness

### `MapUtils` Display Functions
- Renders facility markers with distinct SVG icons
- Handles proper cleanup of markers when needed
- Maps facility types to appropriate visual representations

### SVG icons
- All icons were added to the project. You can see them in the picture below.

## Performance Challenges & Solutions
The initial implementation faced significant performance bottlenecks when directly displaying facilities through the `FacilitiesViewModel`. After careful analysis, we identified two main issues:

1. The original approach attempted to process and display all facilities regardless of the current map state, leading to unnecessary computations and rendering overhead.

2. Without proper filtering based on the map's current view state, the system would try to handle too many facilities simultaneously, causing noticeable lag during map interactions.

To address these challenges, we implemented a new architecture that:

- Tightly couples facility display to the map's current zoom level and visible bounds
- Implements intelligent filtering in the `ViewModel` that considers:
  - Maximum facility counts appropriate for each zoom level
  - Distance-based relevance to the current route
  - Visible map area constraints
- Uses debounced map events to prevent excessive updates during rapid map interactions

This new approach significantly improves performance by ensuring we only process and display facilities that are actually relevant to the user's current view.
image

# Note
I will leave the reviewer decide whether to accept or reject Sonar's problems.


mchac15:
@davidholzwarth 
There is still a problem with two tests that I don't know why are not working. That's why the coverage is low, while with the same tests I had above 80% coverage before. It seems to be a problem with how I handled sensible Thread data in the tests. I'll try to fix it but you can still review the code for the time being to speed up the review, since it doesn't affect at all the functionality itself and because this is a large PR that needs to be merged for friday.

mchac15:
> @davidholzwarth There is still a problem with two tests that I don't know why are not working. That's why the coverage is low, nevertheless I had above 80% coverage with the same tests before. It seems to be a problem with the tests not being Thread safe. I'll try to fix it but you can still review the code for the time being to speed up the review, since it doesn't affect the functionality at all and because this is a large PR that needs to be merged by Friday.

After discussing with the group on telegram and after 7h of trial and error to make the tests work, we decided to keep the coverage a 73% since the project's coverage itself is pretty high already.

app/src/main/res/values/strings.xml
```diff
@@ -227,6 +227,10 @@
 
     <!-- Error handling -->
     <string name="retry">Retry</string>
+
+    <!-- Map Utils -->
+    <string name="facility_marker">facility_marker</string>
```
davidholzwarth, line 232: Nitpick: Since this string never really be translated, it is imo not really necessary to save is as a string resource, its enough to just save it as a component in some object, like Facility or so. But I may be wrong, this side of Kotlin is not my strength.

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -508,4 +567,123 @@ class HikeDetailScreenTest {
...
+        return@runBlocking // Exit successfully
+      }
+
+      delay(delayMs) // Use coroutine delay instead of Thread.sleep
+      attempts++
+    }
+
+    // If we get here, the marker never appeared
+    fail("Marker was not added to map after ${maxAttempts * delayMs}ms")
+  }
```
davidholzwarth, line 647: Praise: Madman, this looks insane! Nice that you got this test working.

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -508,4 +567,123 @@ class HikeDetailScreenTest {
...
+    val facilities: MutableState<List<Facility>?> = mutableStateOf(testFacilities)
+    lateinit var mapView: MapView
+    val minZoomForFacilities = MIN_ZOOM_FOR_FACILITIES
+
+    composeTestRule.setContent {
+      mapView = hikeDetailsMap(detailedHike3, facilitiesViewModel, facilities)
+    }
+
+    composeTestRule.waitForIdle()
+    Thread.sleep(3000)
```
davidholzwarth, line 665: Question: I am scared to touch on this topic haha, but is the sleep here necessary? Isn't waitForIdle enough?

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -508,4 +567,123 @@ class HikeDetailScreenTest {
...
+  @Test
+  fun hikeDetails_hidesFacilities_whenZoomLevelIsInsufficient() = runTest {
+    setUpSelectedHike(detailedHike3)
+
+    val testFacilities =
+        listOf(Facility(type = FacilityType.TOILETS, coordinates = LatLong(45.9, 7.6)))
+
+    val facilities: MutableState<List<Facility>?> = mutableStateOf(testFacilities)
+    lateinit var mapView: MapView
+    val minZoomForFacilities = MIN_ZOOM_FOR_FACILITIES
```
davidholzwarth, line 658: Nitpick: For these kinds of constant, its more readable to include the class where this constant is from, so 
```kotlin
FacilitiesViewModel.MIN_ZOOM_FOR_FACILITIES 
```

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -508,4 +567,123 @@ class HikeDetailScreenTest {
...
+    composeTestRule.waitForIdle()
+    Thread.sleep(3000)
+
+    // Verify facilities are hidden at insufficient zoom levels
+    val finalMarkers =
+        mapView.overlays.filterIsInstance<Marker>().filter {
+          it.relatedObject == R.string.facility_marker
+        }
+
+    assertTrue(
```
davidholzwarth, line 673: Question: Might be a dumb question, but how come the zoom level is actually to far zoomed out in this test? Is it that the hike is "big" enough for this effect to happen?

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -508,4 +567,123 @@ class HikeDetailScreenTest {
...
+        "Facilities should be hidden at zoom level ${mapView.zoomLevelDouble}, " +
+            "below minimum $minZoomForFacilities",
+        finalMarkers.isEmpty())
+  }
+
+  // Helper function to compare drawables
+  private fun areSameDrawable(drawable1: Drawable?, drawable2: Drawable?): Boolean {
```
davidholzwarth, line 680: Praise: sexy

davidholzwarth, line 680: Suggestion: Maybe make this function public? This function seems useful also for other test classes. Perhaps a "base class" from which the two test classes inherit from. There you could also add redundant things like the detailedHikes1, 2 and 3 and maybe even some duplicated mocking behaviors? If this is too much work we could also maybe leave this as a task for another time...

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
```diff
@@ -300,4 +360,159 @@ class RunHikeScreenTest {
...
+    `when`(facilitiesRepository.getFacilities(any(), any(), any())).then {
+      val onSuccess = it.getArgument<(List<Facility>) -> Unit>(1)
+      onSuccess(listFacility)
+    }
+    setupCompleteScreenWithSelected(detailedHike2)
+    verify(facilitiesRepository).getFacilities(any(), any(), any())
+  }
+
+  @Test
+  fun runHike_displaysCorrectDrawableForFacilityType() = runBlocking {
```
davidholzwarth, line 380: Suggestion: Love the test still, but this one two is just copy paste from the hike details, or? It might also be good to inherit from a common base class.

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
```diff
@@ -300,4 +360,159 @@ class RunHikeScreenTest {
...
+      delay(delayMs) // Use coroutine delay instead of Thread.sleep
+      attempts++
+    }
+
+    // If we get here, the marker never appeared
+    fail("Marker was not added to map after ${maxAttempts * delayMs}ms")
+  }
+
+  @Test
+  fun runHike_hidesFacilities_whenZoomLevelIsInsufficient() = runTest {
```
davidholzwarth, line 434: Suggestion: here too, this is also duplicated code, if i see correctly. See the comments about redundancy above.

app/src/main/java/ch/hikemate/app/MainActivity.kt
```diff
@@ -86,6 +89,7 @@ fun HikeMateApp() {
           FirebaseAuthRepository(), ProfileRepositoryFirestore(FirebaseFirestore.getInstance()))
 
   val hikesViewModel: HikesViewModel = viewModel(factory = HikesViewModel.Factory)
+  val facilitiesViewModel = FacilitiesViewModel(FacilitiesRepositoryOverpass(OkHttpClient()))
```
davidholzwarth, line 92: Warning: It seems that the author of the FacilitiesVM (me...) didn't implement a factory, as done with the other vMs. I dont remember exactly why, but there was a clear benefit to doing this. Do you mind adding this to the vM and using that instead? Afaik chatgpt or claude are good at this stuff, or you could just look at e.g. the HikesViewModel.factory

davidholzwarth:
Also, regarding the lower test coverage, since we discussed it in the group, I also do not have a problem with letting this PR pass despite the low test coverage.

app/src/main/java/ch/hikemate/app/model/facilities/FacilitiesViewModel.kt
```diff
@@ -3,17 +3,130 @@ package ch.hikemate.app.model.facilities
...
+   * @param bounds Current visible bounds of the map
+   * @param hikeRoute The detailed hike route being displayed
+   * @return true if facilities should be displayed for the current state, false otherwise
+   */
+  private fun isValidZoomAndBounds(
+      zoomLevel: Double,
+      bounds: BoundingBox,
+      hikeRoute: DetailedHike
+  ): Boolean {
+    // This might seem redundant but the reason this is used is so that if the zoom level
```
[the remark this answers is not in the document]
mchac15: This check is redundant for the ZoomLevel condition. Here's why:
- This function checks whether the zoom is high enough >13
- As you can see in line 65:
 ```kotlin 
val maxFacilities = getMaxFacilitiesForZoom(zoomLevel)
```
This is a "double check" per say because we check if the map contains any facilities whatsoever for that specific `zoomLevel`. That's why it might be deemed "redundant" for zoom levels outside that map.

File: app/src/main/java/ch/hikemate/app/model/route/Hike.kt [hunk 11]
```diff
@@ -143,4 +143,9 @@ data class DetailedHike(
     val estimatedTime: Double,
     val elevationGain: Double,
     val difficulty: HikeDifficulty
-)
+) {
+  /** The list of segments of the Route */
+  val segments: List<RouteSegment> by lazy {
```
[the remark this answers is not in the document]
mchac15: Yes exactly!
This avoids non-useful data being stored
(This was just copied from the HikeRoute.kt for the adaptation so already reviewed)

Not really. If everything goes well and there are no facilities for that state in the designated area then `onNoFacilitiesForState` would not be called. This check just avoids absurd trivial computations.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -245,6 +290,82 @@ fun hikeDetailsMap(hike: DetailedHike): MapView {
...
+          facilitiesViewModel.filterFacilitiesForDisplay(
+              facilities.value ?: emptyList(),
+              bounds = boundingBox,
+              zoomLevel = zoomLevel,
+              hikeRoute = hike,
+              onSuccess = { newFacilities ->
+                // Double check map validity before updates
+                if (mapView.repository != null) {
+                  MapUtils.clearFacilities(mapView)
+                  if (newFacilities.isNotEmpty()) {
```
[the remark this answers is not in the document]
mchac15: Not really. This is the case where all checks pass but there are no facilities for that state. This avoids useless computation.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -245,6 +290,82 @@ fun hikeDetailsMap(hike: DetailedHike): MapView {
...
+              },
+              onNoFacilitiesForState = {
+                if (mapView.repository != null) {
+                  MapUtils.clearFacilities(mapView)
+                }
+              })
+        }
+      }
+    } catch (e: CancellationException) {
+      // Rethrow cancellation exceptions to allow proper coroutine cancellation
```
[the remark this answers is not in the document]
mchac15: `CancellationException` is a special exception in Kotlin coroutines that signals normal cancellation (like when a composable is disposed). I refactored it since in our case it isn't really necessary to throw it, since from what I understand Compose handles coroutines cancellation.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -259,11 +380,20 @@ fun hikeDetailsMap(hike: DetailedHike): MapView {
...
     }
+    // This is useful for the first composition since there won't be any zoom or scroll events at
+    // first so the hike displays the possible facilities from the get go.
+    facilitiesViewModel.filterFacilitiesForDisplay(
+        facilities.value ?: emptyList(),
+        bounds = boundingBoxState.value,
+        zoomLevel = zoomLevelState.value,
+        hikeRoute = hike,
+        onSuccess = { MapUtils.displayFacilities(it, mapView, context) },
+        onNoFacilitiesForState = { /*Nothing since nothing is displayed yet*/})
```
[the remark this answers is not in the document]
mchac15: Not really. Since this is an `addOnFirstLayoutListener` it'll called during the composable execution, so the listener itself will only trigger once.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -245,6 +290,82 @@ fun hikeDetailsMap(hike: DetailedHike): MapView {
...
+        combine(
+            boundingBoxState.debounce(HikeDetailScreen.DEBOUNCE_DURATION),
+            zoomLevelState.debounce(HikeDetailScreen.DEBOUNCE_DURATION)) { boundingBox, zoomLevel ->
+              // Return a pair of our values
+              boundingBox to zoomLevel
+            }
+
+    // Collect the flow with proper error handling
+    try {
+      combinedFlow.collect { (boundingBox, zoomLevel) ->
```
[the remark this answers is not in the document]
mchac15: Yes it is. It keeps collecting until the end of the coroutine execution (CouroutineCancellation). You can test it on your phone when you boot the app.
If it didn't collect further states it wouldn't be displaying the facilities themselfs on the map.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -99,8 +112,32 @@ fun RunHikeScreen(
...
 @Composable
-private fun RunHikeContent(hike: DetailedHike, navigationActions: NavigationActions) {
+private fun RunHikeContent(
+    hike: DetailedHike,
+    navigationActions: NavigationActions,
+    facilitiesViewModel: FacilitiesViewModel
+) {
   // Avoids the app crashing when spamming the back button
+
+  val facilities = remember { mutableStateOf<List<Facility>?>(null) }
```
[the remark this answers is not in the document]
mchac15: I moved the facilities list to the `viewModel`.
The reason I didn't do it before:
I didn't want to adapt the whole `viewModel` to my implementation of the Screens, since this `viewModel` could be used elsewhere. I adapted it by making it still possible to be used as before just with a function more specific to hikes.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -160,6 +201,82 @@ private fun runHikeMap(hike: DetailedHike): MapView {
     }
   }
 
+  // Create state values that we can actually observe in the LaunchedEffect
+  // We keep our StateFlows for debouncing
+  val boundingBoxState = remember { MutableStateFlow(mapView.boundingBox) }
+  val zoomLevelState = remember { MutableStateFlow(mapView.zoomLevelDouble) }
+
+  // Update the map listener to just update the StateFlows
+  mapView.addMapListener(
```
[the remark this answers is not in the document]
mchac15: Done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -160,6 +201,82 @@ private fun runHikeMap(hike: DetailedHike): MapView {
...
+              zoomLevelState.value = newZoomLevel
+            }
+          }
+          return true
+        }
+      })
+
+  LaunchedEffect(Unit) {
+
+    // Create our combined flow
```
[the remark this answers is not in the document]
mchac15: I don't think I agree on this one. This one while redundant in both screens would be difficult to place in an appropriate file.

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -58,8 +74,11 @@ class HikeDetailScreenTest {
   private lateinit var profileViewModel: ProfileViewModel
   private lateinit var mockSavedHikesRepository: SavedHikesRepository
   private lateinit var hikesRepository: HikeRoutesRepository
-  private lateinit var elevationRepository: ElevationRepository
+
```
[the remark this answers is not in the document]
mchac15: No it might be because I missclicked

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -264,4 +268,75 @@ object MapUtils {
...
+   * @param mapView the map view to display the facilities on
+   * @param context
+   */
+  fun displayFacilities(facilities: List<Facility>, mapView: MapView, context: Context) {
+    val displayedFacilities = mutableSetOf<GeoPoint>()
+
+    facilities.forEach { facility ->
+      // Get the drawable corresponding to the facility type
+      val drawable =
+          when (facility.type) {
```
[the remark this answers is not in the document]
mchac15: Done

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
```diff
@@ -79,14 +98,50 @@ class HikeDetailScreenTest {
           estimatedTime = 169.3169307105514,
           difficulty = HikeDifficulty.DIFFICULT,
       )
+  private val detailedHike2 =
```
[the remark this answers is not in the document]
mchac15: Yes the waypoints are different. I added comments to explain it.

app/src/main/res/values/strings.xml
mchac15: Done

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
mchac15: Yes it should be. Actually that's my bad, I should be calling `delay`

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
mchac15: Added

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
mchac15: The bounds are too big.

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
mchac15: The thing is I wouldn't which file would be appropriate for this function. I can make it public for now.

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
mchac15: I don't really understand what you mean by "base class"

app/src/main/java/ch/hikemate/app/MainActivity.kt
mchac15: Done

mchac15:
@davidholzwarth I did a lot of changes and added a bunch of your recommendations. Many things had to change.
Thank you for your patience!

davidholzwarth (changes requested):
Thanks! The new code has like 500 lines less bc of reduced duplication. Could you please still make the two HD and RH screen test classes inherit from a common parent class, to avoid the duplicate test like they are right now?

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
mchac15: I would like to say yes to this. Unfortunately we won't have the time to do it...

mchac15: Anyways I wouldn't know how you would inherit from two test classes and I also see that most tests that handle the screen repeat themselves so we would have to refactor all those tests

app/src/androidTest/java/ch/hikemate/app/ui/map/HikeDetailsScreenTest.kt
davidholzwarth: Perfect!

app/src/androidTest/java/ch/hikemate/app/ui/map/RunHikeScreenTest.kt
davidholzwarth: Ok I agree, there is no time. Let's leave it, I've also discussed with the group. Since its not a critical issue we can leave it.

davidholzwarth (approved):

davidholzwarth (approved):

PR #254, opened 2024-12-08
fix: back handler in hike details not in the right place

rBergCode:
# Summary
This pull request includes changes to the `HikeDetailScreen` in the `HikeDetailsScreen.kt` file to improve the handling of the back button functionality.

Back button handling improvements:
Moving it to the correct place to have it work in all cases.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
@@ -139,6 +139,8 @@ fun HikeDetailScreen(
     }
   }
 
+  BackHandler { hikesViewModel.unselectHike() }
+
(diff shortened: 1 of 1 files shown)


MaelImhof (commented):
Simple and quick fix, thanks! How necessary do you think a regression test would be? I think it [is possible](https://stackoverflow.com/questions/72973956/jetpack-compose-back-button-press-in-automation-test) to trigger a system back button event in the tests.

rBergCode:
I don't think it's relevant here. It's just an implementation error and we don't test it anywhere else

MaelImhof:
True, no need for a regression test, however there are conflicts to be resolved before this can be merged.

MaelImhof (approved):
LGTM, thanks for the merge 👌🏻

PR #259, opened 2024-12-09
fix: ui jump when going from saved to planned

rBergCode:
# Summary
We had a weird UI bug when going from saved to planned hikes. See images:
image
image

# Details
The issues was arising when having more saved than planned hikes. I fixed this by adding a `Column` with `fillMaxSize` in order to make sure the title and list will take as much space as possible

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt
@@ -141,6 +141,9 @@ fun SavedHikesScreen(hikesViewModel: HikesViewModel, navigationActions: Navigati
 @Composable
 private fun PlannedHikes(hikes: List<StateFlow<Hike>>?, hikesViewModel: HikesViewModel) {
   val context = LocalContext.current
+  // Use fillMaxSize() to ensure consistent layout when switching between tabs
+  // This prevents unwanted UI resizing that can occur when sections have different
+  // numbers of items, particularly when transitioning between saved and planned hikes
   Column(modifier = Modifier.fillMaxSize()) {
     Text(
         context.getString(R.string.saved_hikes_screen_planned_section_title),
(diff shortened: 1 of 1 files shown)


app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt
```diff
@@ -141,29 +141,31 @@ fun SavedHikesScreen(hikesViewModel: HikesViewModel, navigationActions: Navigati
...
-          modifier =
-              Modifier.padding(16.dp)
-                  .testTag(SavedHikesScreen.TEST_TAG_SAVED_HIKES_PLANNED_EMPTY_MESSAGE))
-    }
-  } else {
-    LazyColumn {
-      items(plannedHikes.size, key = { plannedHikes[it].value.id }) { index ->
-        val hike by plannedHikes[index].collectAsState()
-        SavedHikeCardFor(hike, hikesViewModel)
+  Column(modifier = Modifier.fillMaxSize()) {
```
mchac15, line 144: I'd suggest adding a comment like this to prevent this bug from happening again with further modifications to this file.
```suggestion
// Use fillMaxSize() to ensure consistent layout when switching between tabs
// This prevents unwanted UI resizing that can occur when sections have different
// numbers of items, particularly when transitioning between saved and planned hikes
  Column(modifier = Modifier.fillMaxSize()) {
```

app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt
```diff
@@ -172,25 +174,28 @@ private fun PlannedHikes(hikes: List<StateFlow<Hike>>?, hikesViewModel: HikesVie
...
-          modifier =
-              Modifier.padding(16.dp)
-                  .testTag(SavedHikesScreen.TEST_TAG_SAVED_HIKES_SAVED_EMPTY_MESSAGE))
-    }
-  } else {
-    LazyColumn {
-      items(savedHikes.size, key = { savedHikes[it].value.id }) { index ->
-        val hike by savedHikes[index].collectAsState()
-        SavedHikeCardFor(hike, hikesViewModel)
+  Column(modifier = Modifier.fillMaxSize()) {
```
mchac15, line 177: Suggestion: Here too:
```suggestion
// Use fillMaxSize() to ensure consistent layout when switching between tabs
// This prevents unwanted UI resizing that can occur when sections have different
// numbers of items, particularly when transitioning between saved and planned hikes
  Column(modifier = Modifier.fillMaxSize()) {
```

mchac15 (changes requested):
Great fix! We just need to document this bug so that it doesn't occur again in the future.
After that you can merge!

app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt
rBergCode: Good idea!

PR #252, opened 2024-12-06
refactor: move all strings to strings.xml

rBergCode:
# Summary
We had strings everywhere in the app. Move them all to the `strings.xml` file.

PR #261, opened 2024-12-09
chore: handle nullable name in saved hike repo

rBergCode:
# Summary
As I was doing #252 , I found out that `SavedHikesRepository` didn't allow nullable names. I changed that in order to have a possibility to have nullable name.


MaelImhof (commented):
LGTM, just one question: does this work with Firebase? In the sense, is it retro-compatible with existing documents and existing accounts that already have saved hikes?

rBergCode:
Yes, as we change it from **non-nullable** to **nullable**, this is not a problem. I tried and everything works perfectly

MaelImhof (approved):
Very nice, looks good to me then!

PR #262, opened 2024-12-09
Fix/bug map too small

davidholzwarth:
# Summary

Whenever scrolling downwards navigating to the south with the map is not shown on the edge of the BottomScaffold. To fix this, I reduced the size of the bottomSheet padding in all screens that have a map, so that the map displays even under the section of the bottomSheet that does not cover the complete width of the screen as a result of its rounded corners.


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -272,7 +272,11 @@ fun hikeDetailsMap(hike: DetailedHike): MapView {
       factory = { mapView },
       modifier =
           Modifier.fillMaxWidth()
-              .padding(bottom = 300.dp) // Reserve space for the scaffold at the bottom
+              .padding(
+                  bottom =
+                      300.dp -
+                          20.dp) // Reserve space for the scaffold at the bottom, -20.dp to avoid
+              // the map being to small under the bottomSheet
```
MaelImhof, line 279: Nitpick: maybe this is a good opportunity to refactor the hike details screen and put those values in constants in the `HikeDetailsScreen` object. The constants should not be shared between screens, as we might want to make the bottom sheet slightly higher on details screen than on map screen for example.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -446,7 +446,11 @@ fun MapScreen(
                     modifier =
                         Modifier.fillMaxSize()
                             .testTag(MapScreen.TEST_TAG_MAP)
-                            .padding(bottom = MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT))
+                            .padding(
+                                bottom =
+                                    MapScreen.BOTTOM_SHEET_SCAFFOLD_MID_HEIGHT -
+                                        20.dp)) // -20.dp to avoid the map being to small under
```
MaelImhof, line 452: Nitpick: define this value in a constant in the `MapScreen` object. The constants should not be shared between screens, as we might want to make the bottom sheet slightly higher on details screen than on map screen for example.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -184,7 +184,11 @@ private fun runHikeMap(hike: DetailedHike): MapView {
       factory = { mapView },
       modifier =
           Modifier.fillMaxWidth()
-              .padding(bottom = 300.dp) // Reserve space for the scaffold at the bottom
+              .padding(
+                  bottom =
+                      300.dp -
+                          20.dp) // Reserve space for the scaffold at the bottom, -20.dp to avoid
+              // the map being to small under the bottomSheet
```
MaelImhof, line 191: Nitpick: maybe this is a good opportunity to refactor the run hike screen and put those values in constants in the `RunHikeScreen` object. The constants should not be shared between screens, as we might want to make the bottom sheet slightly higher on details screen than on map screen for example.

MaelImhof (changes requested):
Nice and clean fix ! Just nitpicking to avoid magic numbers.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: What do you mean? You mean add 20.dp as a constant in the MapScreen object?

MaelImhof:
I believe a test was broken by the new padding, could you take a look at the CI output?

MaelImhof (approved):
Still looks good to me 👌🏻

PR #264, opened 2024-12-09
feat: better disabled style for the search hikes button

rBergCode:
# Summary
As said by the coach, the search button disabled style is not that easy to understand. I changed it to make it look more explicit. I also changed so that when searching hikes, instead of hiding the button, make it disabled.

image
image


davidholzwarth (commented):
Thanks. This solution is very clean! The changes look great. There is no test that verifies that the button is disabled when the zoom level is too high. Could you please add that?

davidholzwarth (changes requested):
Thanks. This solution is very clean! The changes look great. There is no test that verifies that the button is disabled when the zoom level is too high. Could you please add that?

davidholzwarth (approved):
LGTM

PR #270, opened 2024-12-11
fix: adapt Hike.kt and DetailedHike to the projection function

mchac15:
# Summary
Adapted the `LocationUtils`.`projectLocationOnHike` to the new code and the new `DetailedHike` type.
This simple change is necessary for the previous functionality of the function and since the code now accepts `DetailedHike` as a value instead of `HikeRoute`.

### DIFF
--- a/app/src/test/java/ch/hikemate/app/model/route/HikeTest.kt
+++ b/app/src/test/java/ch/hikemate/app/model/route/HikeTest.kt
@@ -169,7 +169,8 @@ class HikeTest {
   @Test
   fun segmentsWorksAsExpected() {
     // Case 1: Empty waypoints list
-    val emptyHike =
+
+    val baseHike =
         createHike(
                 description = "description",
                 bounds = Bounds(0.0, 0.0, 0.0, 0.0),
@@ -180,6 +181,7 @@ class HikeTest {
                 elevationGain = 0.0,
                 difficulty = HikeDifficulty.EASY)
             .withDetailsOrThrow()
+    val emptyHike = baseHike
 
     assertTrue(emptyHike.segments.isEmpty())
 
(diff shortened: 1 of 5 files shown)


app/src/test/java/ch/hikemate/app/model/route/HikeTest.kt
```diff
@@ -165,4 +165,103 @@ class HikeTest {
     assertEquals(elevationGain, detailed.elevationGain, deltaForDoubleComparison)
     assertEquals(difficulty, detailed.difficulty)
   }
+
+  @Test
+  fun segmentsWorksAsExpected() {
+    // Case 1: Empty waypoints list
+    val emptyHike =
+        createHike(
```
MaelImhof, line 173: Praise: thank you for using the existing `createHike` function and keeping the code clean this way!

app/src/test/java/ch/hikemate/app/model/route/HikeTest.kt
```diff
@@ -165,4 +165,103 @@ class HikeTest {
...
+                description = "description",
+                bounds = Bounds(0.0, 0.0, 0.0, 0.0),
+                waypoints = emptyList(),
+                elevation = emptyList(),
+                distance = 0.0,
+                estimatedTime = 0.0,
+                elevationGain = 0.0,
+                difficulty = HikeDifficulty.EASY)
```
MaelImhof, line 181: Nitpick: this is repeated quite a few times, maybe create a `val` at the beginning of the function and use `myVal.copy(waypoints = ...)` to change the waypoints only (because they are the only thing that matters here).

MaelImhof (commented):
Thank you for your quick response on this matter, we realized late we needed this function to be adapted for the new view model too.

Your code is very clean and looks good to me, I just have one nitpick. Thank you for your time on this.

app/src/test/java/ch/hikemate/app/model/route/HikeTest.kt
mchac15: Done

PR #276, opened 2024-12-12
feat: make hikes unplannable

davidholzwarth:
# Summary

This PR adds the option to remove the planned status of a hike by reselecting the date on which the hike is already planned. The pictures below show:

 1. Selecting the date to hike. the button says "Plan this hike"
image

2. date is shown
image

3. Selecting same date again. Now it says "Un-plan this hike"
image

4. The hike is no longer planned. Also in the planned Hikes menu.
image
image


app/src/test/java/ch/hikemate/app/model/route/HikesViewModelTest.kt
```diff
@@ -878,6 +878,42 @@ class HikesViewModelTest {
...
+            hikeId = singleOsmHike1[0].id,
+            onSuccess = { onSuccessCalled = true },
+            onFailure = { fail("onFailure should not have been called") })
+
+        // The saved hikes repository should be called exactly once
+        coVerify(exactly = 1) { savedHikesRepo.removeSavedHike(any()) }
+        // The appropriate callback should be called
+        assertTrue(onSuccessCalled)
+        // The hike should now be marked as unsaved
+        assertNull(hikesViewModel.hikeFlows.value[0].value.plannedDate)
```
MaelImhof, line 914: Suggestion: this does not mean the hike is unsaved, it means it was unplanned, maybe you should check both and adapt comments accordingly?

app/src/test/java/ch/hikemate/app/model/route/HikesViewModelTest.kt
```diff
@@ -878,6 +878,42 @@ class HikesViewModelTest {
...
+  fun `unsaveHike also unplans hike`() =
+      runTest(dispatcher) {
+        // Make sure the loaded hike is included in the saved hikes
+        coEvery { savedHikesRepo.loadSavedHikes() } returns
+            listOf(
+                SavedHike(
+                    id = singleOsmHike1[0].id,
+                    name = singleOsmHike1[0].name ?: "",
+                    date = firstJanuary2024))
+        hikesViewModel.refreshSavedHikesCache()
```
MaelImhof, line 891: Nitpick: you can make use of the util `loadSavedHikes` here. The subsequent `loadOsmHikes(singleOsmHike1)` call you make will override the loaded hikes anyway, so it does not matter here whether you use refresh or load saved hikes.

```suggestion
        loadSavedHikes(listOf(
            SavedHike(
                    id = singleOsmHike1[0].id,
                    name = singleOsmHike1[0].name ?: "",
                    date = firstJanuary2024)))
```

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
MaelImhof, line 1: Question: your implementation works, but I don't think it is user-friendly. Selecting the same hike to unplan the hike is kind of weird. I would say users are more used to something like this :

th-2913643174


With a `Clear` or `Clear date` button on the left. Would that be feasible?

MaelImhof (changes requested):
Nice work! I was wondering if a clear button can be added instead, because selecting the same date to unplan it seems a bit far-fetched for me, at least from a user perspective.

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
davidholzwarth: I hoped so, but the date picker dialog we use (Jetpack Compose's `DatePickerDialog` does not have this option, they only allow a dismiss button (clear) and a confirm button (set). [Here is the documentation](https://composables.com/material3/datepickerdialog). In my opinion, given the time constraints this option works too. Perhaps we could check in the group to see if everyone is ok with this "hack"-y solution?

app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -279,6 +279,8 @@ class HikesViewModel(
    *
    * Setting a null date on an unsaved hike will have no effect.
    *
+   * Setting a null date on a saved hike will removed the planned date but keep the hike saved.
+   *
```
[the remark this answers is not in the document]
davidholzwarth: I removed the documentation in the following lines. Does that work?

app/src/test/java/ch/hikemate/app/model/route/HikesViewModelTest.kt
davidholzwarth: whoops, thats still the comments from when i copy-pasted this test. It now says `The hike should now be marked as unplanned (and unsaved, but that is a previous test)`

davidholzwarth (commented):
About the UX thing, check the comment. We could also ask in the group if thats ok

app/src/test/java/ch/hikemate/app/model/route/HikesViewModelTest.kt
MaelImhof: Perfect 👌🏻

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
MaelImhof: I checked with @rBergCode and my own conscience, and I believe this is good enough for now. We'll see if we have time to improve it (spoiler: probably not), but the option to unplan must exist, even if not ideal.

MaelImhof (approved):
Looks good to me, thank you for your changes. The way the user can unplan a hike is not fully ideal in my opinion, but I agree it might take too much time to improve it, and we have other priorities. We'll see if we have time for this.

MaelImhof:
Also, just to be clear, I did notice the Sonar error, but it's only a cognitive complexity thing, so I marked it as accepted since we have other priorities and don't have much time nor interest to refactor this method.

PR #277, opened 2024-12-13
Test/e2e4

davidholzwarth:
# Summary

This PR adds a fourth end to end test, that checks functionality of the run hike screen. The test currently only checks whether it is possible for the user to navigate to and from the run hike screen, since it is neither feasible to test the facilities on the map nor the location features in an e2e test. In the future, this test should be updated to show whether the data in the bottomsheet details are good, but currently they are still hard coded.


rBergCode:
I put it in draft while we don't have a proper testing of the run hike feature

Wylwi:
Took over by @Wylwi in another PR

PR #278, opened 2024-12-13
feat: increase location updates freq

MaelImhof:
# Summary

When using the run hikes screen, updating the user's location every 5 seconds is a bit slow, does not feel smooth. We empirically observed that a delay of `500` milliseconds is fine in terms of performance and provides a way better user experience in terms of smoothness.


rBergCode (approved):
Thanks!

Wylwi (approved):
LGTM!

PR #282, opened 2024-12-14
chore: fix ui breaking on small and big screens

davidholzwarth:
# Summary

There were multiple UI bugs when using the app on very big or small screen:
1. On small screens, the ButtomSheet in the Map screen covered almost the entire screen
2. On small screens, the "Search hikes here" button covered the zoom buttons partially
3. On big screens, the BottomScaffold did not stretch the entire width of the device, exposing that we do not load the map for the entire screen
4. If the Hike Details screen's bottomSheet was expanded when pressing "Run this hike", when moving from the Run Hike screen back to the Hike Details screen, the bottomSheet would stay expanded up. 
5. The back button in the run hike screen was very low or high, depending no the device, since it still used a hard coded top padding value

@Reviewer Could you please also spend like 10 minutes exploring small and big devices using the Android Studio emulator to see if I missed anything?

# 1
image 
->  
image

The peek height for the map screen was reduced, form 400 to 250dp. The peek height for the other screens stayed at 400 dp, since these do not have a bottom navigation bar, as seen in the image below, here more of the map is visible than in the original version of the map screen

image

# 2 
image 
->  
image

The fix involved loading both buttons and checking for an overlap. If there is an overlap, the "short button" mode is activated, which reduces the length of the text in the button. To compare, on bigger screens the search hikes button text stays "Search hikes here"

image
image

# 3
image -> image

This happened since we did not manually set the BottomSheet's `sheetMaxWidth `. Setting this one to a large number stops the restriction of the width.

# 4

I added a `LaunchedEffect` the the HD screen, that plays a collapsing animation of the bottomSheet if it was still expanded from the last time the screen was quit. Even if the RH's bottomSheet was up when quitting the RH screen (e.g. if the user went back using the go back feature from Android while the bottomSheet is up), this animation plays, which is not ideal but IMO not that bad. This could be fixed by sharing the HD screen's scaffold state among the screens, which would be a bit of a pain to implement, but lmk if that would be better.

# 5

image
->
image


app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -551,7 +578,9 @@ fun MapSearchButton(onClick: () -> Unit, modifier: Modifier = Modifier, enabled:
           ),
       enabled = enabled) {
         Text(
-            text = LocalContext.current.getString(R.string.map_screen_search_button_text),
+            text = if (shortText) "Search here" else "Search for hikes",
+            // text = "Search
+            // here",//LocalContext.current.getString(R.string.map_screen_search_button_text),
```
rBergCode, line 583: Warning: I think you forgot to actually use the `strings.xml` here

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -392,6 +395,17 @@ fun MapScreen(
...
+  var button2Bounds by remember { mutableStateOf<Rect?>(null) }
+  var shortText by remember { mutableStateOf(false) }
+  LaunchedEffect(button1Bounds, button2Bounds) {
+    if (button1Bounds != null &&
+        button2Bounds != null &&
+        button1Bounds!!.overlaps(button2Bounds!!)) {
+      shortText = true
+    }
+  }
+
```
rBergCode, line 408: Question: I understand how that works but what would happen if the device is even smaller so that having a `shortText` is not small enough?

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode, line 1: Warning: I couldn't find how this now happens but I think this PR introduces a new bug. There is a white bar at the top of the MapScreen. It only occurs there.

rBergCode (changes requested):
Good job of finding UI bugs! I've tested on both small and big devices and they now look way better! I have some warnings and remarks and we can merge it after.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: Actually, it looks like it also has this bug on main. You can ignore this for now but please add a task in the SB.

davidholzwarth: Correction: I fixed it, by moving the container the map is in (and thus the top of the map) beyond the screen, so the loading of the map is triggered earlier. 

image

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: It would overlap again.

I agree that its a valid concern, but imo that's not really worth the effort of adding an even smaller button. The small phone from the Android Studio emulator is already extremely small. It would probably be so small that we would also start getting issues with the localisation button. However, if you think its appropriate, I can surely add a third size which just says like "Search" or just has a magnifying glass.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
davidholzwarth: Öhh yes, i forgot to push that. Changed.

davidholzwarth (commented):
Note: I've also renamed a typo `HikesDetailsBottomScaffold` -> `HikeDetailsBottomScaffold` and added a fix for the map screen not loading at the to bug.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -578,7 +589,10 @@ fun MapSearchButton(
           ),
       enabled = enabled) {
         Text(
-            text = if (shortText) "Search here" else "Search for hikes",
+            text =
+                if (shortText)
+                    LocalContext.current.getString(R.string.map_screen_search_button_text_short)
+                else LocalContext.current.getString(R.string.map_screen_search_button_text),
```
rBergCode, line 595: Warning: You should use `stringResources` as in the rest of the code of the app.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: Looks good like this! You don't have to do any more job for this.

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -631,7 +645,7 @@ fun CollapsibleHikesList(
       // Overwrites the device's max sheet width to avoid the bottomSheet not being wide enough
       sheetMaxWidth = Integer.MAX_VALUE.dp,
       sheetContent = {
-        Column(modifier = Modifier.fillMaxSize().testTag(MapScreen.TEST_TAG_HIKES_LIST)) {
+        Column(modifier = Modifier.testTag(MapScreen.TEST_TAG_HIKES_LIST)) {
```
rBergCode, line 648: Warning: I think you meant to remove something else (in HikeDetailsScreen)

rBergCode (changes requested):
Thanks for the changes! One more little change and we can merge this!

rBergCode (approved):
Thanks for the changes! You can merge it now

PR #290, opened 2024-12-14
fix: offline update profile, add timeout to go back.

rBergCode:
# Summary
To reproduce this bug, you have to:
- Sign in as a user
- Disable internet on the device
- Edit the profile (IMPORTANT: you have to actually change it.)
- Save the profile

# Details
I added a delay of **5 seconds** to force the user to go back. This will still update the user in firebase as the task was launched.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
@@ -19,6 +19,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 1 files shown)


Wylwi:
@rBergCode I'm not sure of how this fix the bug? This just emulate the updateProfile behavior?

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -182,6 +190,15 @@ fun EditProfileScreen(
...
                 isLoading = true
+                // Add a timeout in case the user is offline. This would still
+                // update the profile as Firebase Firestore syncs when the user
+                // goes online.
+                coroutineScope.launch {
+                  delay(EditProfileScreen.UPDATE_TIMEOUT)
+                  isLoading = false
+                  savedProfile = null
+                  navigationActions.goBack()
+                }
```
Wylwi, line 201: Warning: I'm not sure of how this works as a Timeout? Imo, this will run no matter what and introduces a new bug in case the user reedit during the 5s timeout?

Wylwi (commented):
Just a few questions about the code! The code works, I've tested it, and it has the expected behavior. I'm just wondering how it works 😅

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
```diff
@@ -93,6 +98,8 @@ fun EditProfileScreen(
 
     var isLoading by remember { mutableStateOf(false) }
 
+    val coroutineScope = rememberCoroutineScope()
```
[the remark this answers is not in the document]
rBergCode: It will create a coroutine scope that will only last as much as the Composable is used.
This would remove the coroutines and not do multiple go backs.

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
rBergCode: Same answer as first one

Wylwi: Ooooh I see! Thanks!

app/src/main/java/ch/hikemate/app/ui/profile/EditProfileScreen.kt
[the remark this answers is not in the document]
Wylwi: Ooooh I see! Thanks!

Wylwi (approved):
Looks clearer to me now! Just wait for the CI to finish and you can merge!

PR #293, opened 2024-12-14
fix(bug): performance and display of facilities on first composition and facilities being drawn under the Route

mchac15:
# Summary
This pull request includes several changes to the `FacilitiesViewModel`, `HikeDetailsScreen`, `RunHikeScreen`, and `MapUtils` classes to improve error handling and to fix an issue where the faciliities are only displayed once the user moves the map.

### Error Handling Improvements:

* Added error logging in the `FacilitiesViewModel` to catch and log exceptions during facility filtering. 

### Map Update Optimization:

* Changed the debounce duration for map updates from 300ms to 500ms to reduce the frequency of refreshes in the `HikeDetailsScreen` class.
* Updated the `boundingBoxState` and `zoomLevelState` to initialize with `null` values instead of the current map view state in both `HikeDetailsScreen` and `RunHikeScreen` classes. 

### Improvements
* Refactored `MapUtils` to include new constants and methods for better readability and performance. 

### Map Overlay Management:
* Ensured that the route overlay is added with the lowest priority so that facilities and other overlays are always displayed on top in `MapUtils`. 
* Improved the `clearFacilities` method to handle concurrent modifications and force garbage collection of markers in `MapUtils`.


app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
```diff
@@ -302,13 +295,13 @@ fun hikeDetailsMap(hike: DetailedHike, facilitiesViewModel: FacilitiesViewModel)
 
   // Create state values that we can actually observe in the LaunchedEffect
   // We keep our StateFlows for debouncing
-  val boundingBoxState = remember { MutableStateFlow(mapView.boundingBox) }
-  val zoomLevelState = remember { MutableStateFlow(mapView.zoomLevelDouble) }
+  val boundingBoxState = remember { MutableStateFlow<BoundingBox?>(null) }
+  val zoomLevelState = remember { MutableStateFlow<Double?>(null) }
```
rBergCode, line 299: Question: I'm not sure why you changed that. Can you explain ?

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -375,13 +371,13 @@ fun runHikeMap(hike: DetailedHike, facilitiesViewModel: FacilitiesViewModel): Ma
 
   // Create state values that we can actually observe in the LaunchedEffect
   // We keep our StateFlows for debouncing
-  val boundingBoxState = remember { MutableStateFlow(mapView.boundingBox) }
-  val zoomLevelState = remember { MutableStateFlow(mapView.zoomLevelDouble) }
+  val boundingBoxState = remember { MutableStateFlow<BoundingBox?>(null) }
+  val zoomLevelState = remember { MutableStateFlow<Double?>(null) }
```
rBergCode, line 377: Question: Same as before, why null here?

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -75,9 +86,11 @@ object MapUtils {
             true
           })
         }
-
-    mapView.overlays.add(line)
+    // The index provides the lowest priority so that the facilities and other overlays
+    // are always displayed on top of it.
+    mapView.overlays.add(ROUTE_PRIORITY_DISPLAY, line)
     mapView.overlays.add(startingMarker)
+    mapView.invalidate()
```
rBergCode, line 93: Question: What does that do? Do we need it ?

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -379,15 +392,24 @@ object MapUtils {
...
+    // Force garbage collection of markers
+    overlaysToRemove.forEach { overlay ->
+      if (overlay is Marker) {
+        overlay.onDetach(mapView)
+      }
     }
 
-    // Trigger the map to be drawn again
+    // Clear the MapView's cache
     mapView.invalidate()
```
rBergCode, line 411: Question: Same, why do we need this part?

rBergCode (changes requested):
Few questions and you will probably be able to merge it soon!

app/src/main/java/ch/hikemate/app/ui/map/HikeDetailsScreen.kt
mchac15: When setting the `BoundingBoxState` and the `ZoomLevelState` to null it is forced to be updated when the screen is launched.
This fixes the bug in some cases where you go from the `RunHikeScreen` to the `HikeDetailsScreen`.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
mchac15: Same as above.

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
mchac15: This forces the map to be redrawn so that the overlays are always **on top** of the Hike Route. This fixed the bug of the facilities being drawn behind the Route. I will update the name of the PR to account for this.

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
mchac15: Same as above. It is to ensure that it is redrawn correctly.

PR #296, opened 2024-12-15
feat: display actual progression and user elevation in run

rBergCode:
# Summary
The current elevation and progress were fake in the run hike screen. This PR changed this to display the actual progression and user elevation.

To get the elevation of the user, I retrieve the elevation of the **starting point of the segment on which the user is projected on**. This is some sort of estimation but is way faster than actually retrieving the elevation of the user from the API.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
@@ -170,14 +170,19 @@ private fun RunHikeContent(
     }
 
     var userLocationMarker: Marker? by remember { mutableStateOf(null) }
+    var completionPercentage: Int? by remember { mutableStateOf(null) }
+    var userElevation: Double? by remember { mutableStateOf(null) }
 
     // We need to keep a reference to the instance of location callback, this way we can unregister
     // it using the same reference, for example when the permission is revoked.
     val locationUpdatedCallback = remember {
       object : LocationCallback() {
         override fun onLocationResult(locationResult: LocationResult) {
-          userLocationMarker =
+          val locationParsed =
               parseLocationUpdate(locationResult, userLocationMarker, mapView, hike)
+          userLocationMarker = locationParsed.first
+          completionPercentage = locationParsed.second
+          userElevation = locationParsed.third
           if (centerMapOnUserPosition &&
               userLocationMarker != null &&
               userLocationMarker?.position != null)
@@ -269,7 +274,11 @@ private fun RunHikeContent(
                   .testTag(RunHikeScreen.TEST_TAG_ZOOM_BUTTONS))
 
       // Display the bottom sheet with the hike details
-      RunHikeBottomSheet(hike = hike, onStopTheRun = { wantToNavigateBack = true })
+      RunHikeBottomSheet(
+          hike = hike,
+          completionPercentage = completionPercentage,
+          userElevation = userElevation,
+          onStopTheRun = { wantToNavigateBack = true })
     }
   }
 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -170,14 +170,19 @@ private fun RunHikeContent(
...
     // it using the same reference, for example when the permission is revoked.
     val locationUpdatedCallback = remember {
       object : LocationCallback() {
         override fun onLocationResult(locationResult: LocationResult) {
-          userLocationMarker =
+          val locationParsed =
               parseLocationUpdate(locationResult, userLocationMarker, mapView, hike)
+          userLocationMarker = locationParsed.first
+          completionPercentage = locationParsed.second
+          userElevation = locationParsed.third
```
MaelImhof, line 185: Question: any reason you did not use the triplet sugared syntax here?

```kotlin
val (val1, val2, val3) = parseLocationUpdate(...)
```

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -314,7 +326,17 @@ private fun parseLocationUpdate(
         }
       }
 
-  return MapUtils.updateUserPosition(userLocationMarker, mapView, newLocation)
+  val marker = MapUtils.updateUserPosition(userLocationMarker, mapView, newLocation)
+  // Progress distance is in meters, hike distance is in kilometers
+  val completionPercentage =
+      if (routeProjectionResponse.distanceFromRoute > RunHikeScreen.MAX_DISTANCE_TO_CONSIDER_HIKE)
+          null
+      else (routeProjectionResponse.progressDistance * 0.1 / hike.distance).toInt()
```
MaelImhof, line 334: Question: why is the `* 0.1` needed here?

app/src/main/java/ch/hikemate/app/utils/LocationUtils.kt
```diff
@@ -245,7 +245,8 @@ object LocationUtils {
           progressDistance = projected.distanceTo(segment.start),
           distanceFromRoute = location.distanceTo(projected),
           segment = segment,
-          indexToSegment = 0)
+          indexToSegment = 0,
+          projectedLocationElevation = route.elevation[0])
```
MaelImhof, line 249: Suggestion: maybe we should handle the case here where `route.elevation` is an empty list.

MaelImhof (commented):
Nice addition to the app! Just a few questions and suggestions, and you're good to go.

app/src/main/java/ch/hikemate/app/utils/LocationUtils.kt
rBergCode: You're right! I changed it to be nullable so that we handle this case!

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
MaelImhof: Fair enough

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
MaelImhof: Right, I did not do the simplification in my head, but that makes sense

MaelImhof (approved):
Thanks for the changes, looks good to me

MaelImhof:
Although make sure the CI passes before merging

MaelImhof (approved):
Still looks good to me, thanks for fixing the tests

PR #298, opened 2024-12-16
fix: fix back buttons double clicks

Wylwi:
# Summary
- This PR simply removes the ability to multi go back by multi clicking on the back button.

# Details
- Clicking too fast on the BackButton would trigger multiple goBack, and raise logical issues

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/components/BackButton.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/BackButton.kt
@@ -12,6 +12,11 @@ import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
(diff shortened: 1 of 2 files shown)


davidholzwarth (approved):
Nice PR. In the Run Hike screen the back button still uses the `wantToNavigateBack` variable. It would be nice to change that too, but either way the PR LGTM

app/src/main/java/ch/hikemate/app/ui/components/BackButton.kt
```diff
@@ -34,8 +39,18 @@ fun BackButton(
     modifier: Modifier = Modifier,
     onClick: () -> Unit = { navigationActions.goBack() }
 ) {
+
+  var wantsToClick by remember { mutableStateOf(false) }
+
+  LaunchedEffect(wantsToClick) {
```
davidholzwarth, line 45: In the Run Hike screen, the `wantToNavigateBack` variable is still used by this button. Just to make you aware, it does not need to be changed but might be nice.

davidholzwarth (approved):

app/src/main/java/ch/hikemate/app/ui/components/BackButton.kt
Wylwi: Actually, the RunHikeScreen has multiple paths that lead to going back, so this makes sense! (Not only the BackButton click)

PR #301, opened 2024-12-17
feat: display the starting of a hike

rBergCode:
# Summary
It was previously very hard to know which end of the hike was the start or the end of it. For this, I added a little circle at the start of it.

# Details
We could discuss if this could still be confusing for the user as he could believe this is the end of the hike.

Before:
before

After:
after

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
@@ -169,6 +169,12 @@ object MapScreen {
    */
   const val USER_LOCATION_MARKER_ICON_SIZE = 40
 
+  /**
+   * (Config) Size of the icon representing the starting point of a hike on the map. The size is
+   * defined empirically to make the icon visible and not too big.
+   */
+  const val HIKE_STARTING_MARKER_ICON_SIZE = 40
+
   // These are the limits of the map. They are defined by the
   // latitude values that the map can display.
   // The latitude goes from -85 to 85, because going beyond
--- a/app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
+++ b/app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
@@ -226,7 +226,7 @@ object MapUtils {
     val bitmap =
         Bitmap.createBitmap(
             MapScreen.HIKE_STARTING_MARKER_ICON_SIZE,
-            MapScreen.USER_LOCATION_MARKER_ICON_SIZE,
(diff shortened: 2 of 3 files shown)


app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
```diff
@@ -169,6 +169,12 @@ object MapScreen {
    */
   const val USER_LOCATION_MARKER_ICON_SIZE = 40
 
+  /**
+   * (Config) Size of the icon representing the starting point of a hike on the map. The size is
+   * defined empirically to make the icon visible and not too big.
+   */
+  const val HIKE_STARTING_MARKER_ICON_SIZE = 40
```
Wylwi, line 176: Question: What unit is this? Is it in pixels? If so, how does it render on small screens?

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -186,13 +206,48 @@ object MapUtils {
...
+   * @return The circular icon with the specified fill and stroke colors
+   */
+  private fun createCircularIcon(context: Context, fillColor: Int): BitmapDrawable {
+    // Create a mutable bitmap
+    val bitmap =
+        Bitmap.createBitmap(
+            MapScreen.HIKE_STARTING_MARKER_ICON_SIZE,
+            MapScreen.USER_LOCATION_MARKER_ICON_SIZE,
+            Bitmap.Config.ARGB_8888)
+    val canvas = Canvas(bitmap)
```
Wylwi, line 231: Question: I don't understand what do this do? Why do you use a **_USER_LOCATION_**_MARKER_ICON_SIZE?

Wylwi (commented):
Overall, your code is clean and good, I've just got a few questions on how it works!

app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
rBergCode: Good catch! my bad

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: I did this the same way as the user location marker. It seems that both render the same in all conditions.
I've just checked on a small phone. We may see it too big but I think it's fine like this. See the image:

image

Wylwi: Mmmmh, yeah, I agree

Wylwi (approved):
LGTM now! Thanks for the small change!

PR #302, opened 2024-12-17
fix: add a back button on HikeDetails and RunHike loading screen

### LINKED ISSUE #299: Fix: offline mode MapScreen -> HikeDetailScreen
Paths to recreate it:
1. Online search for hikes.
2. Disable your internet connection.
3. Try to go to the HikeDetailsScreen of a Hike.
4. There is no back button to go back.

So the user should have the option to go back in this case with the Back Button.

Wylwi:
# Summary

- This fixes #299 where a user could be stuck in a soft loop in case their internet was not working, while trying to go on the HikeDetails screen.
- This does the same thing on the RunHikeScreen, even though the bug should never happen here since we already have the details

# Details
Paths to recreate the bug:

- Online search for hikes.
- Disable your internet connection.
- Try to go to the HikeDetailsScreen of a Hike.
- User is stuck

This now adds a BackButton for the user to escape.

# Old UI

Screenshot_20241217_165010
Screenshot_20241217_165016

# New UI

Screenshot_20241217_164650
Screenshot_20241217_164708

### DIFF
--- a/app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
@@ -57,6 +57,7 @@ fun WithDetailedHike(
 
   if (detailedHike != null) {
     withDetailedHike(detailedHike!!)
+    return
   }
 
   Column(Modifier.fillMaxSize().safeDrawingPadding()) {
--- a/app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
+++ b/app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
@@ -1,17 +1,27 @@
 package ch.hikemate.app.ui.components
 
 import android.util.Log
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.safeDrawingPadding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Refresh
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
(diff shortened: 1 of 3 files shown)


app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
```diff
@@ -37,58 +47,73 @@ fun WithDetailedHike(
...
-    // All the details of the hike have been computed, pass them on to the content callback
-    hike.isFullyLoaded() -> {
-      val detailedHike =
-          try {
-            hike.withDetailsOrThrow()
-          } catch (e: Exception) {
-            Log.d(WithDetailedHike.LOG_TAG, "Error while getting detailed hike: $e")
-            null
-          }
+  var detailedHike: DetailedHike? by remember { mutableStateOf(null) }
```
rBergCode, line 56: Question: Why do you need this variable? It seems it could work without it

app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
```diff
@@ -37,58 +47,73 @@ fun WithDetailedHike(
...
 
-      if (detailedHike != null) {
-        withDetailedHike(detailedHike)
-      } else {
-        whenError()
-      }
+  Column(Modifier.fillMaxSize().safeDrawingPadding()) {
+    if (detailedHike == null && !hike.isFullyLoaded()) {
```
rBergCode, line 59: Same here, I think you only need the second condition

rBergCode (changes requested):
Nice fix! I have a small question and you will be good to go!

app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
Wylwi: Yep, it works without it

app/src/main/java/ch/hikemate/app/ui/components/WithDetailedHike.kt
Wylwi: Yep, it works without it

rBergCode (approved):
Thanks for the changes!

PR #304, opened 2024-12-19
fix(hikes): unsaving hike in details

MaelImhof:
There was a bug when the user went to the saved screen and opened a hike's details. If the user then unsaved the hike, it removed the hike from the loaded hikes, thus unselected then hike, and thus kicked the user out of the details screen.

This bug was "by design" in the sense that the HikesViewModel was built with the idea that when the selected hike is unloaded, it must be unselected as well. But this does not make sense from a practical point of view.

This commit corrects the design of HikesViewModel and its tests to let the user unsave the hike freely.

### DIFF
--- a/app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
+++ b/app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
@@ -87,6 +87,22 @@ class HikesViewModel(
 
   private val _hikeFlowsList = MutableStateFlow<List<StateFlow<Hike>>>(emptyList())
 
+  /**
+   * Indicates whether when unselecting it, the selected hike in [_selectedHike] should also be
+   * removed from [_hikeFlowsMap] (and consequently [_hikeFlowsList]).
+   *
+   * This was introduced to solve a bug where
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/model/route/HikesViewModel.kt
```diff
@@ -547,8 +567,16 @@ class HikesViewModel(
...
+      // Set the flag for removing the hike from the list when it is unselected
+      _selectedHikeShouldBeRemoved = true
+
+      // The selected hike is not in the map, add it back and update its saved status
+      val saved = _savedHikesMap[selectedHike.id]
+      val newValue = selectedHike.copy(isSaved = saved != null, plannedDate = saved?.date)
+      _hikeFlowsMap[selectedHike.id] = MutableStateFlow(newValue)
+      if (selectedHike != newValue) {
+        _selectedHike.value = newValue
+      }
```
rBergCode, line 579: Praise: Clean code here!

rBergCode (approved):
Very clean PR. Nothing to say. Nice fix!

PR #310, opened 2024-12-19
feat:add puck in run hike screen

mchac15:
# Summary
This PR adds the functionality of the location puck of the user being displayed on the `ElevationGraph` in the `RunHikeScreen`.

# Functionality
With the distance from the projection we get the this value to calculate the percentage that has been run to give it to the `ElevationGraph`.


app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -334,11 +342,14 @@ private fun parseLocationUpdate(
   val completionPercentage =
       if (routeProjectionResponse.distanceFromRoute > RunHikeScreen.MAX_DISTANCE_TO_CONSIDER_HIKE)
           null
-      else (routeProjectionResponse.progressDistance * 0.1 / hike.distance).toInt()
+      else
+          (routeProjectionResponse.progressDistance /
+              (hike.distance * RouteUtils.METERS_PER_KIlOMETER))
```
davidholzwarth, line 347: Praise: Nice constant

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -334,11 +342,14 @@ private fun parseLocationUpdate(
   val completionPercentage =
```
davidholzwarth, line 342: Warning: This returns a ratio, not a percentage, please rename or change the return

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -517,19 +528,9 @@ private fun RunHikeBottomSheet(
               modifier = Modifier.testTag(RunHikeScreen.TEST_TAG_HIKE_NAME))
 
           // Elevation graph and the progress details below the graph
+          val hikeColor = Color(hike.color)
+          ElevationGraphOnPercentage(completionPercentage, hike, hikeColor)
```
davidholzwarth, line 532: Question: Why is the new elevation graph not in the column?

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
```diff
@@ -613,3 +616,36 @@ private fun RunHikeBottomSheet(
         }
       }) {}
 }
+
+@Composable
+private fun ElevationGraphOnPercentage(
```
davidholzwarth, line 621: Nitpick: This name is kinda weird. Why is the elevationgraph on the percentage? Maybe something like `ElevationGraphWithLocationPuck` or something like that.

davidholzwarth, line 621: Warning: Is it necessary to have a function for this? The ElevationGraph already handles cases when the completionpercentage is null, there is no need to catch that in this scope, or?

davidholzwarth (changes requested):
Very cool feature, but i think there is some unnecessary redundancy in this code.

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
mchac15: Done

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
mchac15: Big error my part. Sorry. Fixed

app/src/main/java/ch/hikemate/app/ui/map/RunHikeScreen.kt
mchac15: Agreed! 
Fixed

mchac15: Yes because of Sonar complaining about the complexity.

davidholzwarth: If you use an elvis operator, this should work, since ElevationGrapghPercentage has a null check built in

mchac15: Yes you are right. Fixed

davidholzwarth (approved):
LGTM

PR #311, opened 2024-12-19
fix: change the text of the error to be more broad

davidholzwarth:
# Summary

When the device has no account connected, our app launched the account connector process. If this was aborted, the toast that was displayed was still the "success" one, which is misleading to the user. I've made the toast more broad, to cover both success and failure. 

This PR does not fix the bug that when aborting this process, sometimes the app gets stuck on the Google sign in page. This is because we do not have any control over Google's account connection intent. I was thinking maybe we can do a sort of timeout, but that would have to be ultra long, so that it does not cancel the activity while the user is just taking their time logging in (typing password manually, needs to go look for google password, etc.), so ultimately it would be quite useless.


rBergCode (dismissed):
I did some fixes but it's good now!

rBergCode (approved):
Perfect!

MaelImhof (approved):
LGTM

PR #312, opened 2024-12-19
fix: display planned hikes dates

MaelImhof:
# Summary

Quick PR to fix two bugs I noticed lately.

- The planned date was not displayed anymore in the saved hikes screen in the planned hikes section
- `HikeCard` did not display its message before it could display the elevation graph. For the `MapScreen`, it kind of made sense to do this, but not for the planned hikes screen, so I changed it.

# Details

- Added the displaying of planned dates just like it [originally did it](https://github.com/HikeMate/hikeMateApp/blob/937189ea84218093821bf9c0836ea052b8f58b2f/app/src/main/java/ch/hikemate/app/ui/saved/SavedHikesScreen.kt#L107)
- Added a regression test for the displaying of the planned dates
- Made the `HikeCard` display its message if needed even if the elevation is not available
- Updated `MapScreen` accordingly to avoid displaying a message before it was available


app/src/main/java/ch/hikemate/app/ui/theme/Color.kt
```diff
@@ -3,6 +3,7 @@ package ch.hikemate.app.ui.theme
 import androidx.compose.ui.graphics.Color
 
 val primaryColor = Color(0xFF3B9DE8)
+val plannedDateMessageColor = Color(0xFF3B82F6)
```
Wylwi, line 6: Question: Why don't you use the already defined colors?

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -73,7 +73,10 @@ fun HikeCard(
     styleProperties: HikeCardStyleProperties = HikeCardStyleProperties(),
     showGraph: Boolean = true,
 ) {
-  val displayMessage = !elevationData.isNullOrEmpty() && messageContent != null
+  val displayMessage =
+      messageContent != null &&
+          styleProperties.messageColor != null &&
+          styleProperties.messageIcon != null
   Log.i(
```
Wylwi, line 80: Praise: Great fix! It doesn't make sense to have the elevation in it!

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -73,7 +73,10 @@ fun HikeCard(
     styleProperties: HikeCardStyleProperties = HikeCardStyleProperties(),
     showGraph: Boolean = true,
 ) {
-  val displayMessage = !elevationData.isNullOrEmpty() && messageContent != null
+  val displayMessage =
+      messageContent != null &&
+          styleProperties.messageColor != null &&
+          styleProperties.messageIcon != null
```
Wylwi, line 79: Question: Why do we have to enforce the `messageColor` and `messageIcon` to be non-null to display the message? Can't we display it even without it?

If you choose to keep it, could you document it?

app/src/androidTest/java/ch/hikemate/app/ui/saved/SavedHikesScreenTest.kt
```diff
@@ -360,4 +361,30 @@ class SavedHikesScreenTest : TestCase() {
...
+        // Select the planned hikes tab
+        composeTestRule
+            .onNodeWithTag(
+                SavedHikesScreen.TEST_TAG_SAVED_HIKES_TABS_MENU_ITEM_PREFIX +
+                    SavedHikesSection.Planned.name)
+            .performClick()
+        composeTestRule.waitForIdle()
+
+        composeTestRule
+            .onAllNodesWithTag(HikeCard.TEST_TAG_IS_SUITABLE_TEXT, useUnmergedTree = true)
```
Wylwi, line 387: Question: What does `useUnmergedTree` do? Is it mandatory?

Wylwi (commented):
Great bug fixes! Just a few questions about some unusual things!

app/src/main/java/ch/hikemate/app/ui/theme/Color.kt
MaelImhof: Because it is not the same color, and my task did not involve changing the color itself, so I kept it as-is.

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
MaelImhof: One needs the three of those to actually display the message. The user can pass a null message and valid icon/color, or a null icon and a valid message/color, and so on.

So I enforce we have all three to make sure we can actually display the message. Does that sound reasonable?

app/src/androidTest/java/ch/hikemate/app/ui/saved/SavedHikesScreenTest.kt
MaelImhof: In my understanding (I must say I'm not sure to know perfectly what it is), `useUnmergedTree = true` tells `onAllNodesWithTag` to look for nested components. Think of it as a UI hierarchy type of thing. In our case there is the `HikeCard` component, and **inside** of it you have the actual message. I want to find the message, so I use `useUnmergedTree`.

What I know 100% is that without it the test does not pass.

app/src/main/java/ch/hikemate/app/ui/theme/Color.kt
Wylwi: Can you please use the primaryColor instead?

app/src/androidTest/java/ch/hikemate/app/ui/saved/SavedHikesScreenTest.kt
Wylwi: If the test doesn't pass without it, that's good for me!

Wylwi (changes requested):
Please see above comments for changes

app/src/main/java/ch/hikemate/app/ui/theme/Color.kt
MaelImhof: Done 👌🏻

app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
MaelImhof: Done 👌🏻

Wylwi (approved):
LGTM now! Thanks!

PR #314, opened 2024-12-19
fix: ui glitching when clicking on already selected bottom item

Wylwi:
# Summary
- Fixes a UI Glitch bug when clicking on currently selected bottom navigation item 

# Details
- When the user clicked on the current selected bottom navigation item, the navigation actions are triggered and re-render the screen, which leads to a UI glitch. 
- I fixed it by making the navigation actions be triggered only when needed

# Old behavior

https://github.com/user-attachments/assets/f785d91e-7142-4dc1-8856-7208c11a3e9e

# New behavior

https://github.com/user-attachments/assets/204b735a-1497-440f-9f9e-63a3bd5b2816

### DIFF
--- a/app/src/androidTest/java/ch/hikemate/app/navigation/BottomBarNavigationTest.kt
+++ b/app/src/androidTest/java/ch/hikemate/app/navigation/BottomBarNavigationTest.kt
@@ -1,0 +1,84 @@
+package ch.hikemate.app.navigation
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import ch.hikemate.app.ui.navigation.BottomBarNavigation
+import ch.hikemate.app.ui.navigation.LIST_TOP_LEVEL_DESTINATIONS
+import ch.hikemate.app.ui.navigation.Route
+import ch.hikemate.app.ui.navigation.TEST_TAG_BOTTOM_BAR
+import ch.hikemate.app.ui.navigation.TEST_TAG_MENU_ITEM_PREFIX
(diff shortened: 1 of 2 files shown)


app/src/main/java/ch/hikemate/app/ui/navigation/BottomBarNavigation.kt
```diff
@@ -44,8 +45,8 @@ fun BottomBarNavigation(
                   )
                 },
                 label = { Text(stringResource(tab.textId)) },
-                selected = tab.route == selectedItem,
-                onClick = { onTabSelect(tab) },
+                selected = selected,
+                onClick = { if (!selected) onTabSelect(tab) },
```
MaelImhof, line 49: Praise: very elegant fix 👌🏻

app/src/androidTest/java/ch/hikemate/app/navigation/BottomBarNavigationTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class BottomBarNavigationTest {
+  // Set up the Compose test rule
+  @get:Rule val composeTestRule = createComposeRule()
+
+  @Test
+  fun clickingOnAItemChangeScreen() {
```
MaelImhof, line 26: Nitpick: I believe there's an `n` and a `s` missing here

```suggestion
  fun clickingOnAnItemChangesScreen() {
```

MaelImhof (changes requested):
Very elegant PR, thanks for the fix! Only a quick nitpick because I don't have anything else to say.

app/src/androidTest/java/ch/hikemate/app/navigation/BottomBarNavigationTest.kt
Wylwi: Good catch!!

PR #317, opened 2024-12-19
chore/better-ui-ux

rBergCode:
# Summary
We built our app part-by-part. In result, the whole app has different fonts, colors, design choices, etc.
To fix this, I revised the whole app in order to use `MaterialTheme` and `Colors.kt` values. Here is the complete list of fixes:
- Changed the design of the graph to make them prettier
- Using the safe padding everywhere and have the same spaces everywhere
- Move all the fonts in the `MaterialTheme`. This includes applying correct font to some text missing it
- Move all the colors in Color.kt
- Make the run hike and details screen more related to the figma design
- Make the bottom sheet in hike details and run screen smaller by default and stop at the correct height
- Changed "Intermediate" to "Amateur"

# Details
There is a lot of changes. Feel free to contact me to discuss any questions you might have on it.


app/src/main/java/ch/hikemate/app/ui/components/HikeCard.kt
```diff
@@ -110,7 +108,7 @@ fun HikeCard(
                                 fillColor =
                                     (styleProperties.graphColor
                                             ?: MaterialTheme.colorScheme.primary)
-                                        .copy(0.1f)))
+                                        .copy(0.5f)))
```
mchac15, line 119: Praise: I like this change! It looks way better

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
mchac15, line 1: Question: Is it not necessary to add the top padding here? Or it is not possible because of the map?

mchac15, line 1: Bug: I found  a kind of weird which I have no idea of how to fix. The thing is that when you go up the screen with the bottom sheet there is a little black spacing between the BottomBarMenu and the BottomSheet

mchac15 (changes requested):
Great change!
This changes the app completely and feels different.
Just a couple changes and you can merge

app/src/main/java/ch/hikemate/app/ui/components/CenteredLoadingAnimation.kt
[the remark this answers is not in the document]
rBergCode: Good catch!

app/src/main/java/ch/hikemate/app/ui/components/BigButton.kt
```diff
@@ -25,8 +26,8 @@ import ch.hikemate.app.ui.theme.primaryColor
  * @property border The border of the button.
  */
 enum class ButtonType(val backgroundColor: Color, val textColor: Color, val border: BorderStroke) {
-  PRIMARY(primaryColor, Color.White, BorderStroke(0.dp, Color.Transparent)),
-  SECONDARY(Color.White, Color.Black, BorderStroke(1.dp, Color.Black)),
+  PRIMARY(primaryColor, onPrimaryColor, BorderStroke(0.dp, Color.Transparent)),
```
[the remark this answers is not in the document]
rBergCode: I moved it to the display of the button. I can't really do more than that as `.border` doesn't accept `null`

app/src/main/java/ch/hikemate/app/ui/map/MapScreen.kt
rBergCode: Good catch!

mchac15: Thanks

mchac15: Good fix

mchac15 (changes requested):
I still have a bug while navigating through the map the `MapView` below appears with a weird black line. I think you are handling the padding below incorrectly.
This happens both in the Map and HikeDetails screens

mchac15 (changes requested):
I also see that the color used for the _Stop the run_ button in the `RunHikeScreen` in darkmode is kind of weird.
Could you change it please?

mchac15 (approved):
Thanks a lot for the changes!
LGTM

PR #321, opened 2024-12-19
feat: add password requirements hint

davidholzwarth:
# Summary

- When creating an account with a password thats too short, the sign up would fail, but not tell you why. Now it tells you that the password is too short.


rBergCode (approved):
Perfect! Thanks!

PR #322, opened 2024-12-19
fix(bug): operations in the HikeDetailsScreen show a toast when there is no internet connection

mchac15:
# Summary
This PR adds a timeout to operations that are related to the `HikeDetailsScreen` with the `SavedHikesRepository` so that, whenever there is no internet connection the operations timeout and a Toast is shown on the screen.


app/src/main/res/values/strings.xml
```diff
@@ -235,4 +235,5 @@
 
     <string name="delete_account_button_label">Delete Account</string>
     <string name="google_account_created_confirmation">Connected Google Account to your device successfully. Please wait while we retry the signup.</string>
+    <string name="generic_error_message_internet_connection">An error occurred. Check your internet connection</string>
```
davidholzwarth, line 238: Change: Here, perhaps warn the user that the behavior is now undefined. The change may be saved or not, we can not provide a guarantee for either.

davidholzwarth (changes requested):
Please change the error string

app/src/main/res/values/strings.xml
mchac15: Fixed

davidholzwarth (approved):
LGTM

PR #326, opened 2024-12-19
fix: fix unplanning label in hike details

MaelImhof:
# Summary

There was a bug in the details screen where the user would save a hike, plan it, then unsave it, save it again and try to plan it, but the button would read "Unplan this hike" instead of "Plan this hike" when there would be no planned date.

This is a fix for this bug.


mchac15 (approved):
LGTM

PR #330, opened 2024-12-19
fix: appearance of the hike start marker

mchac15:
# Summary 
Add a proper location SVG which designates the start of the Hike.
In the `MapUtils.showHikeOnMap` function added a parameter `withMarker` which designates whether to show the marker or not.


app/src/main/java/ch/hikemate/app/utils/MapUtils.kt
```diff
@@ -70,26 +72,27 @@ object MapUtils {
...
     }
 
-    val startingMarker =
-        Marker(mapView).apply {
-          // Dynamically create the custom icon
-          icon =
-              createCircularIcon(
-                  context = mapView.context,
-                  fillColor = color,
-              )
```
rBergCode, line 80: Warning: You didn't remove the `createCircularIcon`. Please remove the function as we won't use it anymore

rBergCode (changes requested):
Nice addition! Please remove the function as said in my comment before merging

rBergCode (approved):
Thanks!

PR #334, opened 2024-12-19
fix: remove unused-data in xml

rBergCode:
# Summary
We had un-used data in `Color.xml` and `Strings.xml`. This PR removes them.


Wylwi (approved):
LGTM!

PR #336, opened 2024-12-19
fix: profile soft loop bug

Wylwi:
If the user experiences an error with something related to the profileViewModel, this error is never cleared, leading to a soft loop. This PR fixes it


rBergCode (approved):
Nice fix! Thanks

